Issue #48, opened 2016-07-17
Improve location detection / save 


bagage:
Currently when one opens http://brouter.de/brouter-web, he always get map centered on Frankfurt with zoom level 8 or so, the first thing to do is to move the map over his own position. 
It could be great to save the last used position so that when you always bike in the same area, you do not need to re-localize yourself again. I know that permalink can solve this issue, but I'd like to know if a cookie could be saved to store position for default position (like openstreetmap.org does for instance, but also google maps)?
I'd be glad to make a PR for this if you think it's worth it! 

edit: a "localize me" button could be nice too, see [this demo](https://www.domoritz.de/leaflet-locatecontrol/demo/).

- [x] : use `Leaflet.RestoreView` plugin to save and restore last map view. Cf PR #49 
- [x] : add a `Localize me` button using `leaflet-locatecontrol` plugin. Cf PR #49

nrenner:
It seems like there is a plugin [Leaflet.RestoreView](https://github.com/makinacorpus/Leaflet.RestoreView) for keeping the last map position.

Should be straightforward to add, I can do that, not really a need for a PR. But generally, contributions are welcome!

Using the [leaflet-locatecontrol](https://github.com/domoritz/leaflet-locatecontrol) you linked is already on my to-do list together with some basic mobile support.

bagage:
So this issue could be split up in two:
- [x] : use `Leaflet.RestoreView` plugin to save and restore last map view. Cf PR #49 
- [ ] : add a `Localize me` button using `leaflet-locatecontrol` plugin.
Basic mobile support would be awesome indeed when you must check routes while traveling :). Looking forward for it too!

PR #49, opened 2016-07-20
Use leaflet.restoreview.js plugin to save and restore last user's map…

bagage:
… position between sessions (#48)

I am not JS developer so please tell me if something is wrong with my PR. Thanks @nrenner!


bagage:
Added another patch for 2nd part of #48. I added reference to Awesome font in `index.html`, but maybe you wish not to use it?
Again any comment's welcome, I am a truly newbie here!

nrenner:
Thanks for the PR! I do have comments, but don't worry, still learning myself. And I should have more documentation.

The build setup using Bower and Gulp is still a bit rough and not ideal. Also already outdated again, as people seem to have moved on to Webpack nowadays. But what I'm trying to achieve is that all dependencies are referenced and downloaded through Bower, so that the origin and the version used are documented and that managing and updating the depencencies gets a bit easier.

**leaflet.restoreview:**

I would prefer ``js/plugin/leaflet.restoreview.js`` to be added and downloaded via Bower although its an overhead in this case. The dir name ``plugin`` is a bit misleading here, intended for own configuration and extension of external plugins only. 

As it has no bower.json and is not registered with bower (https://bower.io/search/), I think you can reference the repo and add it to the bower.json with 

    bower install makinacorpus/Leaflet.RestoreView#master --save

Then add the file to include (leaflet.restoreview.js) to the "overrides" section. Check if included with ``gulp debug``.

**leaflet.locatecontrol**

Yes, I would prefer not to load another fontset, as we already have [Glyphicons](http://getbootstrap.com/components/#glyphicons) from Bootstrap, but Glyphicons is a bit limited and seems not to have a spinner. Not sure about the marker icon, maybe [fa-location-arrow](http://fontawesome.io/icon/location-arrow/) or something similar to ``glyphicon glyphicon-screenshot`` (but more GPSy) instead, what do you think?

But I might as well merge the commit as it is and think about that later, as I want to add more icons anyway and am not decided on the icon-/fontset yet.

bagage:
Thanks! 

- [x] I moved leaflet.restoreview to bower 
- [x] I removed Awesome font dependency and instead use `glyphicon-refresh` and `glyphicon-screenshot` icons instead. They are not perfect but I guess it's OK for now. I also added overrides for `locatecontrol` plugin but I am unsure about it; could you take a look at it please?

nrenner:
Thanks for the update. Yes, the icons will do for now.

The bower override for locatecontrol isn't necessary, as the project already has a bower.json with a proper [main section](https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/bower.json#L9-L12).

Could you please remove that again, then I think we're ready to merge.

bagage:
OK I removed it, I can't wait to use these on brouter-web! :smile:

Issue #41, opened 2016-01-04
toggle drawing mode (disable/enable) via panel button not keyboard button


KarlaF90:
Hi,

I love BRouter and I love its web interface!! As it is very helpful to use it via smartphone it would be really great to be able to toggle between disabled and enabled drawing mode via a small hitbox in the panel. As you cannot use the hotkeys 'q' and 'd' on the touchscreen this is really 	beneficial. This would make BRouter Web usable on Android and iPhone. 

Together with issue #40 (toggle information panels) or with a resizeable panel BRouter Webs usability on smartphones would significantly increase.

Thank you for your nice service :+1:

nrenner:
Thanks. Already thought that having a button to toggle the drawing mode would be useful. Also see my [comment](https://github.com/nrenner/brouter-web/issues/40#issuecomment-169766805) to #40.

KarlaF90:
Lieber Norbert, 
Meinst Du, dass es für das Brouter web Interface noch zu Beginn der RadSaison das Toggle Tooltip Feature geben wird? Dann könnte man das extrem gute Brouter auch mit dem Iphone benutzen. Das wäre mega praktisch. 
Beste Grüße, 
Karla
    


-------- Ursprüngliche Nachricht --------
Von: Norbert Renner <notifications@github.com> 
Datum: 07.01.2016  19:46  (GMT+01:00) 
An: nrenner/brouter-web <brouter-web@noreply.github.com> 
Cc: KarlaF90 <Bike90@ok.de> 
Betreff: Re: [brouter-web] toggle drawing mode (disable/enable) via panel
  button not keyboard button (#41) 

Thanks. Already thought that having a button to toggle the drawing mode would be useful. Also see my comment to #40.

bagage:
Hello @nrenner, 
I am quite interested in mobile support and I guess this issue is the easiest to solve following my #49 PR, so if noone is working on it I might suggest a PR for it? Let me know ;-).

nrenner:
Sure, go ahead. I started looking into mobile support, but this is independent.

See the delete route button using the EasyButton plugin as an example. You should be able to call ``routing.draw([true|false])``.

nrenner:
Implemented by @bagage with #50.

PR #50, opened 2016-07-28
Toggle drawing mode via panel button (#41)

bagage:
I used `glyphicon-road` instead of `glyphicon-pencil` because it looks more fun - but I am ok to change it obviously!


bagage:
brouter

nrenner:
Awesome!

Issue #57, opened 2016-11-28
style.css modifications are not tracked by gulp watch


bagage:
When modifying the `style.css` file, I have to run `gulp` manually to get it properly generated even if `gulp watch` is running.
I can see in `gulpfile.js` various mentions to `styles` folder but it is commented: should it be working?

nrenner:
Sorry, somehow the notification for this issue got lost.

Watching CSS hasn't been implemented yet. I have been running ``gulp styles`` manually, and when tweaking CSS I sometimes edit directly in Chrome and copy over when done.

PR #58, opened 2016-11-30
New gulp debug task and watch CSS folder

bagage:
Here is a fix for #57 as well as a new target `gulp debug` so that files do not get uglified when developing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,21 +60,24 @@ This is needed for pre-loading the selected profile (unless you allowed local fi
 
 ## Build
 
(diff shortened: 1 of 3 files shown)


nrenner:
Thanks. Don't know how I came to think an additional watch would be more complicated, still had watching CSS on my todo list.

I sometimes use the ``inject`` task to insert script tags when debugging doesn't work properly, but then you have to remember not to commit and to revert manually, so the debug option is surely more convenient.

Issue #66, opened 2017-01-30
PR #56 follow-up


nrenner:
Collection of remaining issues after PR #56:

- [x] Chrome flex/layout issue (map height stays same) when minimizing+maximizing elevation diagram - @bagage 
- [x] Verify Leaflet 1.0 support #65 - @nrenner  
- [x] Big touch icons in Chrome - #69 - @nrenner  
  Combination of changed `L.Browser.touch` feature detect in Leaflet 1.0 and Chrome now supporting touch features on desktop - will create a separate issue
- [x] [Hover/mouseOver not working on Chrome #67](https://github.com/MrMufflon/Leaflet.Elevation/issues/67) - @nrenner
  same as above, because `L.Browser.touch` is now set in Chrome, Leaflet.Elevation [only listens](https://github.com/MrMufflon/Leaflet.Elevation/blob/05d24f4659e5d109d372c869a355ad2a900d7e79/src/L.Control.Elevation.js#L101-L116) for touch events
- [x] now with `dist` folder removed from repo, provide a convenient way for end users to run BRouter-Web offline and update readme [Install instructions](https://github.com/nrenner/brouter-web#install-brouter-web-client) accordingly
  - upload built zip with dist as release (tag) attachment?
    - complete standalone zip including BRouter server jar, profiles and [scripts](https://github.com/abrensch/brouter/tree/master/misc/scripts/standalone) (maintain in brouter-web)?
  - Leaflet commits dist folder only for release/tag, see [build/publish.sh](https://github.com/Leaflet/Leaflet/blob/703ae02aa8cbd0b87be5b01e77754b83ad732267/build/publish.sh#L21) (--force to add ignored)
  - Dockerfile?
- [x] Keep Permalink or implement Hash URL (#62) (in addition to "Share URL")  
  Otherwise users need to switch twice: first Permalink to Share URL, later to Hash URL;  
  Share URL is a visible alternative & allows easy copying, but lacks some features the Permalink has @bagage (see #74)
- [x] Custom profile in sidebar instead of Options modal (at least on desktop) @bagage (see #76)
  - [x] Separate profile and data table in sidebar - #90
- [x] Verify deleted CSS - #92
- [x] [Restore (short) attribution on the map #110](https://github.com/nrenner/brouter-web/issues/110)
- [x] [Remove unused code #111](https://github.com/nrenner/brouter-web/issues/111)
[...]

Issue #34, opened 2015-10-02
Basic mobile support


nrenner:
Touch support for controls (stopPropagation for touch events?), see reported issues:

[brouter web: Pulldown menu do not work](https://groups.google.com/d/msg/osm-android-bikerouting/knPyGLZO-Rw/gex4v5WMBgAJ)
- Profile and Alternative option lists not selectable

[Is there any internet browser for android devices which works with http://brouter.de/brouter-web/ (altitude profile, 4 possibilities, GPX download)?](https://groups.google.com/d/msg/osm-android-bikerouting/ZGjjeD-IpLI/mdF3SSZn2AkJ)
- not able to insert (copy and paste) Profile code
- not able to click/touch on GPX download

bagage:
I know that you did some work regarding this issue @nrenner (and maybe #35), can you share the current status please? I'd again gladly help to make it (and #35 too) happens!

nrenner:
Was going to ask and share what I have. I would be happy if you would take over the mobile support!

Actually, there isn't much yet, just some first experiments, see the `mobile-test` branch. I also have some notes, I will clean these up and make a wiki page tomorrow.

What are your ideas? What would be your approach?

nrenner:
@bagage See wiki pages [Mobile](https://github.com/nrenner/brouter-web/wiki/Mobile) and [Mobile code snippets](https://github.com/nrenner/brouter-web/wiki/Mobile-code-snippets). So I guess that's all I have right now.

bagage:
Thanks, awesome! Your wiki seems complete enough of existing issues; I have yet no solution as a developer, but I'll dig into some code & your references & other responsive design maps to be inspired from.
I'll update you when I have more precise ideas.

RoPP:
[Here](https://github.com/RoPP/brouter-web) attempt to make BRouter-web a little bit more mobile friendly. 
This is still a WIP branch and there is still some missing features but it should be enough to test and see if it is the way to go.

![screen shot 2016-08-09 at 11 49 05](https://cloud.githubusercontent.com/assets/45775/17512954/a3a35502-5e2a-11e6-8036-6e6e86f3d578.png)

nrenner:
Wow, that's already a working mobile version!

Some thoughts:
- Profile name: consider long names e.g. 'vm-forum-velomobil-schnell'
- Elevation graph: What was the reason for having our own implementation? What are your plans for that?
[...]

PR #76, opened 2017-04-25
Move custom profile in sidebar

bagage:
Cf #66 todolist.


nrenner:
Thanks for this so far. I should have explained more. What I meant was to add the profile as a separate tab (or whatever) like it was before, so that both profile and data table can then each on their own be shown in full screen height.

Some background: I keep postponing it, but my first goal for BRouter Web was to better support profile development and customization. Part of this would be providing some sort of IDE/editor using [CodeMirror](http://codemirror.net/) or similar, e.g. like [overpass turbo](http://overpass-turbo.eu/) or [Tangram Play](https://mapzen.com/tangram/play/).

Furthermore, being able to edit the routing profile at runtime is a distinctive feature of BRouter and I think it should be shown prominently. Also, people are often discussing what tags routers are supporting or which are preferred - another reason to raise awareness that to some extent they can see for themselves by looking at the profiles and for BRouter Web to provide an easy way to view and experiment with profiles. So to me the profile input field is more than just as small box to paste some profile text.

As for using tabs like before or something else, I increasingly think that using [sidebar-v2](https://github.com/Turbo87/sidebar-v2) instead might be an idea. Maybe on the right, with the layer switcher included, similar to osm.org, and maybe in flat instead of floating design and/or with a transparent bar background. But that might be for later.

nrenner:
Addition to clarify:
All this is assuming that when no custom profile is active, the currently selected profile is shown. This is not working with brouter.damsy.net because of forbidden cross-origin requests and also needs to be [configured](https://github.com/nrenner/brouter-web/blob/master/config.js#L56-L61) when running locally.

bagage:
I reconfigured properly (I think) my server. I am still using the default brouter instance, but now profiles are hosted on brouter.damsy.net so that they can be fetched. So the clarification issue is fixed :).
Still, I wasn't aware of your goals here: I'll redesign this PR a bit to better fit this purpose.

nrenner:
> Still, I wasn't aware of your goals here

I briefly mentioned it in a [previous comment](https://github.com/nrenner/brouter-web/issues/34#issuecomment-245409552):

> [...] some other collapsible Sidebar will also do for the data table and for editing the profile.
> My original motivation for brouter-web was to build some kind of IDE for profile development and a planning tool with a focus on desktop.

The need for a different layout is also quite obvious when showing the currently selected profile, but I see how this feature can be overlooked, therefore the clarification.

nrenner:
While you're at it, a related issue is scrolling, see #86.

nrenner:
How do we go on with this PR? Shall we merge the current state or would you want to continue updating it?

bagage:
Please merge as is and create a separate issue for the redesign? I think it will be a major update so I prefer to split it in 2 separate PRs.

Issue #73, opened 2017-03-18
Integrate OpenMapSurfer


Ph1r3ph0x:
Maybe we can integrate OpenMapSurfer. It offers the best rendering I know of...
I attempted an integration here:
https://github.com/Ph1r3ph0x/brouter-web/commit/44c4075bbd84c6dc82aa2fa59d4237f041fec2d2
but wasn't able to test...

nrenner:
The low zooms of OpenMapSurfer are indeed very nice. 

But the layer doesn't seem to be updated any more: "OSM-Data update: 09/05/2016". Therefore I guess I wouldn't want to integrate it.

To make your integration work, you would need to add it to the [baseLayers  list](https://github.com/nrenner/brouter-web/blob/master/js/Map.js#L67-L74) of the layer switcher and run the [build](https://github.com/nrenner/brouter-web#build).

In case you are running your own local [installation](https://github.com/nrenner/brouter-web#installation), additional layers can also be added in the [config.js](https://github.com/nrenner/brouter-web/blob/master/config.js#L70-L74) (no build necessary, just a page reload).

bagage:
@nrenner I believe this issue can be closed since #77 allows users to add their own tiles services?

joker234:
Since the OpenMapSurfer tiles are not available anymore you can close this issue. See also:
* https://github.com/leaflet-extras/leaflet-providers/pull/354#issuecomment-627505860
* https://ask.openrouteservice.org/t/mapsurfer-discontinued/1873

bagage:
Thanks for the info @joker234 :)

PR #77, opened 2017-05-02
Allow user to add custom layers

bagage:
This would solve #73 but also my issue, eg. a French (super)great TMS which is for "single user" access only, so we cannot add it to Brouter but as a user I can :).

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -141,6 +141,23 @@
         </div>
     </div>
 
+    <!-- Layers modal window -->
+    <div class="modal fade" id="custom_layers" tabindex="-1" role="dialog" aria-labelledby="Layers window" aria-hidden="true">
+        <div class="modal-dialog modal-lg" role="document">
+                    <h4 class="modal-title">Customize layers</h4>
+                </div>
+                <div class="modal-body">
+                    <input class="form-control" type="text" id="layer_name" spellcheck="true" wrap="off" placeholder="Custom layer name. (ex: OpenStreetMap)"></input>
+                    <input class="form-control" type="text" id="layer_url" spellcheck="false" wrap="off" placeholder="Custom layer URL. (ex: https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png)"></input>
+
+                    <button type="button" id="custom_layers_add_base" class="btn btn-success">Add base layer</button>
+                    <button type="button" id="custom_layers_add_overlay" class="btn btn-success">Add overlay</button>
+                    <button type="button" id="custom_layers_remove" class="btn btn-danger">Remove selection</button>
+                    <table id="custom_layers_table"></table>
+                </div>
+            </div>
+        </div>
+    </div>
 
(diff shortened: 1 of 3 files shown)


krisanselmo:
Nice feature, thanks! 
The use of strava heatmap overlay may be very useful (according to me) for hike or ride routing.
[https://wiki.openstreetmap.org/wiki/Strava](https://wiki.openstreetmap.org/wiki/Strava)
I can't wait to test :)

bagage:
@krisanselmo you can give it a shot on http://brouter.damsy.net 
Any feedback appreciated!

nrenner:
Very nice addition! Indeed useful for regional, personal-use-only (German agencies also have these) and other not included maps.

Not relevant for merging, but what do you think about adding the button inside the layer switcher (e.g. just a [quick hack](http://playground-leaflet.rhcloud.com/fole/edit?html,output))? I feel we should rather reduce the number of buttons on the map if not needed in daily use.

If you wanted, I guess it might even make sense to turn this into a [Leaflet plugin](http://leafletjs.com/plugins.html#develop-your-own).

nrenner:
@krisanselmo I added the Strava heatmap to my todo list of potential layers to add. Would need to check the terms of use. I'm generally open for map suggestions, but would want to avoid having too many layers in the list.

bagage:
> Not relevant for merging, but what do you think about adding the button inside the layer switcher (e.g. just a quick hack)? I feel we should rather reduce the number of buttons on the map if not needed in daily use.
Yeah sure it sounds good. Small fix first, the button should have `type=button` to [avoid submitting POST on click](http://stackoverflow.com/questions/2825856/html-button-to-not-submit-form). But besides that, the proposed hack cannot work as is because [leaflet is iterating over `input` children](https://github.com/Leaflet/Leaflet/blob/cbaf02034c916e0e3ea1f1f5c21d08c41efd0b3e/src/control/Control.Layers.js#L393-L402). Do you know how to workaround this too @nrenner?

nrenner:
> Small fix first, the button should have type=button to avoid submitting POST on click.

Thanks for the hint.

> But besides that, the proposed hack cannot work as is because leaflet is iterating over input children.

Good catch, but would it matter if it stays `button type=button`, instead of `input type=button`? Or maybe using a styled link (`a href="#"`, with [`preventDefault`](http://leafletjs.com/reference-1.0.3.html#domevent-preventdefault) or `stop`), see Bootstrap [Button tags](http://getbootstrap.com/css/#buttons-tags).

Generally the hack was really just that, and more for me to see if adding a button would work at all and to show what I meant, not how to do it. There probably are better ways, perhaps none of them ideal. I haven't looked deeper, just saw that `layerSwitcher.getContainer().innerHTML += '...'` (or better `appendChild`) didn't work right away, because the button would still be shown when collapsed.

nrenner:
Merged this now and created a separate issue for moving the button into the layer switcher: #89

PR #85, opened 2017-05-05
feat: add a gulp command for release

### LINKED ISSUE #67: latest commit (23.1.2017) breaks it for me
naked html startup without any active contents, bot firefox + chrome

all o.k. with commit 1e26cb1027e6144db34669a3621dc15a43c67360 (12.12.2016)

RoPP:
Usage: `gulp release --tag {major, minor, patch} --token <github_token>`

This task will bump version number in package.json, bower.json and index.html. Then create new commit and new tag, push it to origin/master. Last but not least, create zip file with production files and make a github release with zip attached.

This should fix #67.


bagage:
@nrenner: do you think `.zip` release file should contain `key.js` and `config.js`, or should it contain `key.js.template` and `config.js.template`?
Since it's user's own onfiguration I would say the latter, but…? (Currently, this PR does the former).

nrenner:
It should not contain the `key.js`, only the template, as I consider this private and it runs without.

It can/should contain the `config.js`. Not sure about a template there, it would help to avoid overwriting an existing file, but would be an additional step for getting new users started.

bagage:
User will probably have to tweak `config.js` a little (or at least give it a look); I personally would prefer to have it named `config.js.template` (or else) to avoid overwriting as you stated. Up to you what you think is the most important (ease of updating vs ease of creating new instance).

nrenner:
Then let's switch to `config.js.template`.

nrenner:
Wow, that's the deluxe version, thanks!

Issue #60, opened 2016-12-22
Cannot use localization with Chrome50+ in http mode


bagage:
Cf [this](https://www.mapsmarker.com/faqs/geolocaton-https-only/). We might instead disable the button on such browers then? Reported in #59.

![screen shot 2016-12-22 at 09 06 20](https://cloud.githubusercontent.com/assets/24696260/21420494/fdb3dd5e-c825-11e6-9f1d-cd1922c727e4.png)

davidmartinezros:
Hi @bagage!

I have the same problem.

And I don't want to pay for a https domain.

Any clue for using navigation.localization? Can I use any other service that don't need https mode?

bagage:
@davidmartinezros I really don't know - I don't think proper workarounds exist, but you can acquire HTTPS certificate for free now using [letsencrypt](https://letsencrypt.org/). 

By the way @nrenner do you think it could be added on brouter.de website?

nrenner:
> We might instead disable the button on such browers then?

Yes, already thought it might make sense to either disable or hide the geolocation button, maybe with an explaining tooltip (on mobile?) message. Or at least have a more user-friendly error message. Ideally implemented by the [Locate plugin](https://github.com/domoritz/leaflet-locatecontrol) with a cusomizable tooltip/message?

Not sure if there is a proper way to detect this limitation though, the [Google article](https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only#can_i_detect_at_runtime_if_the_geolocation_was_blocked_because_of_not_being_on_a_secure_context) only suggests parsing the error message (which might differ across browsers).

> [...] but you can acquire HTTPS certificate for free now using letsencrypt.
> By the way @nrenner do you think it could be added on brouter.de website?

@abrensch?

bagage:
@abrensch Do you think you can set it up for brouter.de? Ideally redirecting all traffic from http to https would be even better. Configuring https with letsencrypt is quite easy, if you need assistance on it I can help.

nrenner:
The locate control is now only added in a secure context, as this is now required by most Browsers (row "Secure context required"):
https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API#Browser_compatibility

PR #87, opened 2017-05-08
Use https scheme whenever possible, to avoid mixed content issues

bagage:
To fix security issues like #60, the only proper solution will be to switch to HTTPS support. I setup it on https://brouter.damsy.net - there might be other issues but here's a first patch :).

### DIFF
--- a/js/Map.js
+++ b/js/Map.js
@@ -8,28 +8,28 @@ BR.Map = {
 
         var maxZoom = 19;
 
-        var osm = L.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
+        var osm = L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
             maxZoom: maxZoom
         });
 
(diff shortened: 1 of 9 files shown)


js/Map.js
```diff
@@ -8,47 +8,47 @@ BR.Map = {
...
             maxZoom: maxZoom
         });
-        var outdoors = L.tileLayer('http://{s}.tile.thunderforest.com/outdoors/{z}/{x}/{y}.png' + thunderforestAuth, {
+        var outdoors = L.tileLayer('https://{s}.tile.thunderforest.com/outdoors/{z}/{x}/{y}.png' + thunderforestAuth, {
             maxNativeZoom: 18,
             maxZoom: maxZoom
         });
 
-        var esri = L.tileLayer('https://{s}.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}', {
+        var esri = L.tileLayer('http://{s}.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}', {
```
nrenner: Shouldn't this stay https?

bagage: Yes it should - somewhat screwed my test on this one, thanks...

PR #132, opened 2018-09-01
Add leaflet strava segments (work in progress)

bagage:
I've developed a leaflet plugin to display [Strava segments](https://support.strava.com/hc/en-us/articles/216918167-Strava-Segments) on the map. Apart of helping Strava users find segments near them, it allows to find interesting ways around you even for non-Strava users:

image

If Strava API public is not set, buttons won't be added on the interface. WDYT?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,13 +24,15 @@
     "seiyria-bootstrap-slider": "^9.8.1",
     "url-search-params": "~0.5.0",
     "Leaflet.RestoreView": "makinacorpus/Leaflet.RestoreView#master",
+    "leaflet.stravasegments": "bagage/leaflet.stravasegments#master",
     "leaflet.locatecontrol": "^0.60.0",
     "font-awesome": "^4.7.0",
     "bootstrap-select": "hugdx/bootstrap-select#patch-1",
     "leaflet-sidebar-v2": "nrenner/leaflet-sidebar-v2#dev",
     "leaflet.editable": "^1.1.0",
     "codemirror": "^5.35.0",
-    "mapbbcode": "MapBBCode/mapbbcode#v1.2.0"
+    "mapbbcode": "MapBBCode/mapbbcode#v1.2.0",
+    "@mapbox/polyline": "^0.2.0"
   },
   "overrides": {
     "leaflet": {
--- a/config.template.js
+++ b/config.template.js
@@ -125,4 +125,7 @@
         ];
 
     }
+
+    // Strava API token in case you want to display Strava segments
+    BR.conf.stravaToken = null;
 })();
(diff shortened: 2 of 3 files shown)


nrenner:
Nice, I can see how this could be useful, although less so for me personally.

It would be nice to be able to add the segments as an overlay to the layer switcher. This would allow to hide already loaded segments and maybe to only show the buttons when the layer is activated. A simple solution for the plugin might be to add the features to a FeatureGroup instead of the map directly and to offer a `getLayer()` method. The rest could be done in the application.

Haven't fully read and understood the [Strava Brand Guidelines](https://developers.strava.com/guidelines/) (linked from [Developers](https://developers.strava.com/)), but it seems the plugin would need to attribute and link to Strava in a certain way.

I'm fine with merging this now and will probably take care of registering an API key and stuff later. Let me know if/when this is ready to merge.

bagage:
Hey @nrenner, thanks for the "review", I've created bug for the next release about them as indeed they must be addressed at some point.

In the mean time, the plugin can be merged as is, and I've deployed it to brouter.damsy.net for now :).

saesh:
Hey @bagage! Appreciate the work! Strava segments are useful for me.

Unfortunately the new `bower` dependency gives be an error when building `brouter-web`:

    bower @mapbox/polyline=@mapbox/polyline#^0.2.0         EINVALID Name must be lowercase, can contain digits, dots, dashes, "@" or spaces

Any ideas?

I am using:

   - `bower 1.8.4`
   - `node 8.12.0`

bagage:
Cool, I'm glad that you find it interesting too :).

I'm not an node/bower expert but it sounds strange. I'm running similar versions (1.8.4 and 8.11.2) and it works ok. Could you try cloning brouter-web from scratch and retry (`npm install && bower install`) to see if it works?

saesh:
I have found out that it is some weird behavior when not using the cache.

Do this and is should fail:

- `bower cache clean`
- `bower install`

And to make it work again (building the cache for `bower` packages):

- `bower install @mapbox/polyline`
- `bower install`

Now I have to find a way to make this work in my automated build environment, where there is no cache. 🤔

nrenner:
I actually have the same issue.

The npm name has changed from "polyline" to "@mapbox/polyline" after v0.2.0, which is still the latest tagged release. The bower repository still has the old name, https://bower.io/search/ for "@mapbox/polyline" only lists the name "polyline".

Changing "@mapbox/polyline" to "polyline" in `bower.json` seems to work. But I haven't registered a Strava account yet, can you please test if Strava segments work for you with that change?

saesh:
Thanks @nrenner, this did the trick!

I have added a build step to replace the package name:

    sed -i.original 's/\@mapbox\/polyline/polyline/g' bower.json

nrenner:
I just pushed the fix.

bagage:
Thanks @nrenner, it's odd I hadn't the issue here.. glad to see it sorted out :)!

Issue #144, opened 2018-12-08
Support nogo polygons


Phyks:
Hi,

BRouter supports having any polygon (and polyline) as a nogo area instead of only circles since https://github.com/abrensch/brouter/pull/97. I think it would be super useful to have it in BRouter-web, first as users might be interested into having more control than just circles, but also as BRouter-web is a super nice interface to test and debug quickly BRouter code (so it would be super useful for devs :).

I know UMap (http://umap.openstreetmap.fr/, https://github.com/umap-project/umap/) lets user draw simply polygons on a Leaflet map, so maybe part of this code could be reused? (sorry, I don't know much about brouter-web internals :/)

FWIW, my typical use case for having polygons and polylines (beside dev) is that many cities around here are providing OpenData feeds of current road works. These could be loaded somehow in BRouter-web to avoid crossing these areas.

Thanks!

nrenner:
From #157:
> The possibility to draw and edit the polygons as well as their weight. I'm not sure I can help a lot on this, I don't know much about the Leaflet plugin currently used

Drawing with [Leaflet.Editable](https://github.com/Leaflet/Leaflet.Editable) should work with the `start*` [methods](http://leaflet.github.io/Leaflet.Editable/doc/api.html#editable-method) within a [Leaflet.EasyButton](https://github.com/CliffCloud/Leaflet.EasyButton) similar to:
https://github.com/nrenner/brouter-web/blob/70a0f73d12fde7c4972427bfc475874f9a0ba52c/js/plugin/NogoAreas.js#L56

Editing should work automatically [on click](https://github.com/nrenner/brouter-web/blob/70a0f73d12fde7c4972427bfc475874f9a0ba52c/js/plugin/NogoAreas.js#L36-L39) on every layer.

As mentioned above, I added a [custom delete button](https://github.com/nrenner/brouter-web/blob/70a0f73d12fde7c4972427bfc475874f9a0ba52c/js/plugin/NogoAreas.js#L302-L421) that would need to be adapted to polylines and polygons.

What kind of input do you imagine for the weight? An input field for a float cost value, a slider, a dropdown with categories?

Phyks:
> Editing should work automatically on click on every layer.

Indeed, I just made a minor fix and this is working.

> What kind of input do you imagine for the weight? An input field for a float cost value, a slider, a dropdown with categories?

[...]

PR #148, opened 2019-01-04
Start support of nogos polylines/polygons

Phyks:
Hi,

Here is a very first step towards fixing #144. Nogos polylines and polygons should be supported through URL parameters. This also adds support for the weights for circle nogos.

Left to do after this PR:

- [ ] Support custom weights for polylines/polygons nogos (but I think this will require a matching PR in Brouter itself).
- [ ] Do the UI counterpart of this. I'll galdy leave this to anyone more experimented than me with Leaflet, if anyone is willing to handle it :)

Best,


nrenner:
Thanks! Will have a look after the weekend.

Phyks:
Thanks! Concerning the support of custom weight for polylines/polygons, I created a PR upstream here https://github.com/abrensch/brouter/pull/138 and this can be easily added to brouter-web once the PR is merged upstream. :)

PR #150, opened 2019-01-24
Fix release script

bagage:
Regression since 7f4a0bf, the zip file has been renamed but it is not uploaded to github automatically anymore.


nrenner:
Thanks for the fix. Looks like I made an inconsistent change missing the `release:zip` task.

Issue #146, opened 2018-12-12
Optional layers tree


nrenner:
A separate list of additional map layers, organized in a hierarchical tree, that can be added to the layer switcher by the user, see #145.

- [x] check existing plugins: [Layer switching controls](https://leafletjs.com/plugins.html#layer-switching-controls)
- [x] reference/use existing [tile server collections](https://wiki.openstreetmap.org/wiki/Community_Hub#Map_Layer.2FTiles_Collections)

Tree components:
- [jonmiles/bootstrap-treeview: Tree View for Twitter Bootstrap](https://github.com/jonmiles/bootstrap-treeview)
- [maxazan/jquery-treegrid: TreeGrid plugin for jQuery](https://github.com/maxazan/jquery-treegrid/)

Further work after #164:

- [x] persist with local storage
- ~~Strava, Bing, DigitalGlobe~~ (later)
- [x] hash: use id?, ~~unchecked layer~~
- [x] country/language prefixes, ordering
- [x] i18n, ~~JOSM layer translations?~~ (later)
- [x] attribution, map link

nrenner:
Further work after #164:

- [ ] persist with local storage
- [ ] Strava, Bing, DigitalGlobe
- [ ] hash: use id?, unchecked layer
- [ ] country/language prefixes, ordering
- [ ] i18n, JOSM layer translations?
- [ ] attribution, map link
- [ ] ...

Issue #145, opened 2018-12-12
OpenCycleMap and Outdoors alternatives


nrenner:
I got a notice from Thunderforest that our tile usage (about 1,340,000 tile requests per month) is exceeding the free quota and that we need to upgrade to the "Solo Developer" plan, which is £95 (107 €) per month.

I'm aware that running a tile server is demanding and expensive, but at the moment I'm neither willing nor able to pay such an amount, especially as BRouter-Web is non-commercial and not making any money.

So I'm looking for alternative tile services with world-wide coverage and current data to add to the layer switcher, that can divert the traffic or replace the [OpenCycleMap](https://wiki.openstreetmap.org/wiki/OpenCycleMap) and [Outdoors](https://thunderforest.com/maps/outdoors/) layers.

The idea is to also provide a separate list, where the user can choose additional layers from. These can also be regional/specific maps that are only of interest for part of the users. Below a list of candidates I found.

@ all: Do you know some more?

### Commercial providers

Out of interest I did a a pricing comparison of some commercial tile services. None of them seems to include bicycle/hiking routes in their outdoor styles.

Provider | Free tiles¹ | Our price¹² | Pricing¹² | Outdoor styles 
-------- | ---------- | --------- | ------- | ---- 
[Thunderforest](https://thunderforest.com/pricing/) | 150,000 | 107 € | 1,500,000 tiles =  £95 | [OpenCycleMap](https://thunderforest.com/maps/opencyclemap/), [Outdoors](https://thunderforest.com/maps/outdoors/) 
[Mapbox](https://www.mapbox.com/pricing/) | 750,000³ | 18 € ⁵ | $0.50 per 1,000 map views⁴ > 50,000 = 20 $<br><sup>(1340000/15-50000)/1000*0.50</sup> | [Outdoors](https://www.mapbox.com/maps/outdoors/)
[MapTiler Cloud](https://www.maptiler.com/cloud/plans/) | 100,000 | 55 € | 20 $ + $0.05 per 1000 tiles > 500,000 = 62 $ <br><sup>20+((1340000-500000)/1000*0.05)</sup> | [Topo](https://www.maptiler.com/maps/#topo//raster/10.51/7.9234/46.6122)
[...]

Issue #97, opened 2017-11-12
Host custom configs server-side?


bagage:
On brouter.damsy.net I'd like to add some profiles (mainly MTB and trekking-no-path). 
But since I use brouter engine from brouter.de, it seems I must also use the profiles located on it (eg http://brouter.de/brouter/profiles2/).

@nrenner is there any way to host some "custom" profiles server-side?

poutnikl:
Including custom profiles to the brouter-web menu offering is definitely possible in local brouter-web installation. That was the main reason for my experimenting with local brouter-web server. 

The config.js file of the brrouter-web folder explicitly enumerates the profiles for the menu. The profile presence in profiles2 subfolder is not enough. Al least this was true about year ago. 

I guess if such a modified config.file is provided among the server files..........

nrenner:
> But since I use brouter engine from brouter.de, it seems I must also use the profiles located on it (eg http://brouter.de/brouter/profiles2/).

Yes, the engine can only load profiles from that directory and uploaded custom session profiles.

> @nrenner is there any way to host some "custom" profiles server-side?

Currently not. There is the idea to be able to make uploaded custom profiles more permanent for sharing and bookmarking (and potentially adding to local dropdown), see related issues #9 and #26. But that would not change the publicly available profile list.

As @poutnikl said, the current solution/workaround for having a custom set of profiles is to install your own server instance.

bagage:
OK thanks to both :). I'm currently developing an extension to brouter-web to support extra profiles - more to come later.

bagage:
For information I started some WIP in my branch: https://github.com/bagage/brouter-web/commits/dev 
Basically when user selects a custom server profile, brouter-web automatically uploads it and switch to custom profile. It's not an ideal solution but at least it requires minor changes to be effective and up on demo server brouter.damsy.net
Among bugs and improvements:
- an error popup appears temporary between selection of custom server profile and the time of profile is uploaded on server. This popup should not appears.
- instead of changing profile to "custom" we should stick to the profile selected instead, since user did not mean to use a custom profile…

nrenner:
Thanks for the info. Glad you found a solution.

AntoineLu:
Best would actually be to be able to upload a profile from a URL. This way, we could store our profile on our personal github for example and then just upload it from a "raw" url to the file.

Issue #63, opened 2017-01-12
Translations: make website localizable (i18n)


bagage:
Currently the website is only in English. Would be great if it was localized, for instance by using [Leaflet.i18n](https://github.com/yohanboniface/Leaflet.i18n) or [gulp-html-i18n](https://www.npmjs.com/package/gulp-html-i18n).

nrenner:
A lot of [OSM-related projects](https://wiki.openstreetmap.org/wiki/Translation#Editors_and_other_contributing_tools) use [Transifex](http://transifex.com/) for translation, which seems to be free for Open Source projects. There is also an [OpenStreetMap group](https://www.transifex.com/openstreetmap/public/), but mostly with editing-related projects.

nrenner:
- [[OSM-dev] Reminder: OpenStreetMap Organisation on transifex](https://lists.openstreetmap.org/pipermail/dev/2017-June/029920.html)
- [Replace update_locales with Transifex client usage by simon04 · Pull Request #408 · tyrasd/overpass-turbo](https://github.com/tyrasd/overpass-turbo/pull/408/files)

bagage:
OK so I've quickly looked around and in my opinion:

* either we go for the overpass approach, copy/pasting their code. Pros: should be quick and easy. Cons: not standard at all.
* othewise, I propose to drop bower+gulp in favor of npm+webpack, and use [webpack-i18n](https://github.com/webpack-contrib/i18n-webpack-plugin) for instance. Pros: (hopefully) simplify build system and remove deprecated gulp/bower couple. Cons: will take more time for sure. I've no experience with webpack but I guess it's a good occasion to learn :).

What do you think @nrenner or anybody else?

nrenner:
There are two parts to this:
1. synchronizing translation files with Transifex
2. the actual translation lookup function

Both need to work together and agree on the translation file layout and structure.

I only noted the overpass-turbo PR as a current example of a [Transifex CLI Client](https://docs.transifex.com/client/introduction) integration. But there are other examples using Transifex, see "at Transifex" links on the [Translation wiki page](https://wiki.openstreetmap.org/wiki/Translation) or the [Transifex OpenstreetMap group](https://www.transifex.com/openstreetmap/public/). Don't know what they use or if Transifex is offering or recommending a JS library for the 2. part.

I'm all for using standard libraries, but don't know if webpack-i18n really is what people use, there seems to be little documentation, so i wonder:
- will it work with Transifex formats?
[...]

PR #152, opened 2019-01-24
Add strava layer in overlays

bagage:
So that user can show/hide at will.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -76,6 +76,12 @@
       ],
       "dependencies": null
     },
+    "leaflet.stravasegments": {
+      "main": [
+        "dist/leaflet-triangle-marker.js",
+        "dist/L.StravaSegments.js"
+      ]      
+    },
     "leaflet-control-geocoder": {
       "main": [
         "dist/Control.Geocoder.js",
--- a/js/index.js
+++ b/js/index.js
@@ -241,11 +241,12 @@
         nogos.preventRoutePointOnCreate(routing);
 
         if (BR.keys.strava) {
-            new L.Control.StravaSegments(
+            const stravaControl = new L.Control.StravaSegments(
             {
                 stravaToken: BR.keys.strava
             })
             .addTo(map);
+            layersControl.addOverlay(stravaControl.stravaLayer, "Strava segments");
         }
 
         map.addControl(new BR.OpacitySlider({


nrenner:
Thanks, the plugin improvements seem to address my concerns, planning to add a key with the next release.

bagage:
That's great! I plan to review my work branch to see what are the remaining diff between our hosted versions, merge last pieces and probably shut it down.

* The main difference I see right now is #87, eg add HTTPS support on brouter.de. 
* Also some custom profiles I've added (trekking-no-path, mtb and mtb-avoid-road) that are not available on brouter.de instance, but maybe they can be added directly? Or it'll need work on custom profiles (#97 #24)
* The profiles are displayed a bit differently, but this is not a major feature and can be dropped if needed.

nrenner:
> Also some custom profiles I've added (trekking-no-path, mtb and mtb-avoid-road) that are not available on brouter.de instance, but maybe they can be added directly?

Please also open an issue for those to add on abrensch/brouter.

I'm a bit sceptical about offering MTB profiles as riding single-trails in forests is illegal in the part of Germany where I live.

> The profiles are displayed a bit differently, but this is not a major feature and can be dropped if needed.

I think what we really want here is making the profile list translatable. I wanted to wait with translations until the app is more feature-complete and mature, but maybe we could register and setup Transifex and start with translating the profile names.

My current priority is offering additional tile layers though (#145 #146). Improving custom profile handling hopefully next.

bagage:
> Please also open an issue for those to add on abrensch/brouter.
OK, did it in https://github.com/abrensch/brouter/issues/141

> I think what we really want here is making the profile list translatable. [...]
Sounds good. Then my "hack" is not needed anymore. Did you investigate i18n yet since https://github.com/nrenner/brouter-web/issues/63? If you know a way to do it, I might invest some time in it.

nrenner:
No, I haven't. Just collected [these two links](https://github.com/nrenner/brouter-web/issues/63#issuecomment-457679565) on the topic.

nrenner:
I can look into applying for a free account on Transifex and maybe becoming a member of the OpenStreetMap group.

nrenner:
>> I think what we really want here is making the profile list translatable. [...]

> Sounds good. Then my "hack" is not needed anymore.

From a quick look it seems Transifex is using a primary language (English in our case) as the lookup key for the translate function. Don't know how this will work with the profile list, we might still need some sort of mapping from the current profile name to the English display name.

PR #154, opened 2019-02-10
Remove bower

### LINKED ISSUE #116: Replace Bower with Yarn/npm
Bower is deprecated:
> ...psst! While Bower is maintained, we recommend using Yarn and Webpack for front-end projects read how to migrate! 

https://bower.io/

And projects are dropping bower support:
- [Bootstrap](https://getbootstrap.com/docs/4.0/migration/#beta-2-changes)
  > Dropped Bower support as the package manager has been deprecated for alternatives (e.g., Yarn or npm). See bower/bower#2298 for details.
- [leaflet.pm](https://github.com/codeofsumit/leaflet.pm#install-via-bower-deprecated---leafletpm-is-no-longer-supporting-bower-since-0170)
  > Install via Bower (DEPRECATED - leaflet.pm is no longer supporting bower since 0.17.0)

Migration:
- [How to migrate away from Bower? · Bower blog](https://bower.io/blog/2017/how-to-migrate-away-from-bower/)
  > Indeed, until recently Yarn neither could install Bower packages (i.e. GitHub repositories without package.json) nor resolve semver ranges on git tags. I focused on this for a while and the result is pleasing: Yarn 1.x is able to install most of Bower packages.
- what about npm?
- check which dependencies are without package.json

bagage:
Closes #116 

Related to #63, in first place I'd like to remove bower. Still WIP…

Non resolved issues yet:

- [x] get ride of browserMainFiles and regenerate same gulp paths without it (done via `npmFiles`)
- [x] handle deps without package.json, eg only `nrenner/leaflet-routing` https://github.com/nrenner/leaflet-routing/issues/1
- [x] images are searched in `brouter-web` instead of `brouter-web/dist` --> it looks like there is some internal magic happening in gulp.js, @nrenner any clue? To solve the issue for now for dev purpose, we can do `ln -sr dist/{images,fonts} .`.
- [x] issue with async: `async.js:1120 Uncaught TypeError: Cannot set property 'async' of undefined`.
  For now, I have forced jquery and async to be loaded first. But this seems quite ugly and error prone. Do you see any better alternative there (except from using webpack or similar, which would be a much bigger step I believe)?


bagage:
@nrenner so I think getting ride of bower should be quite easy as soon as:

- a package.json is added to leaflet-routing dependency
- the `dist` prefix issues for images/css loading is fixed.

Regarding dependency loading order, I believe we cannot handle it much better with npmFiles/gulp, so maybe we can stick with current workaround (it looks OK for now) and go to webpack or similar for proper dependency management at a latter time?

nrenner:
> * a package.json is added to leaflet-routing dependency

[Done](https://github.com/nrenner/leaflet-routing/commit/bbe5df8af500948251337a7e6c09d14f9fad6d8d).

> * the `dist` prefix issue for images/css loading is fixed.

I pushed a [commit](https://github.com/nrenner/brouter-web/commit/43357eee0d95aaf98f44fbd178201a4858d2b963) from my local branch, where I recently improved the CSS URL rebasing in the gulp `styles` task, which seems to fix the issue.

For it to work, I needed to remove the `datatables/media/css` directory in `paths.styles` by escaping the backslash in RegExp: `RegExp('.*\\.css', 'i').test(f) &&`.

> Regarding dependency loading order, I believe we cannot handle it much better with npmFiles/gulp, [...]

I have also found [main-yarn-files](https://github.com/aimad-majdou/main-yarn-files), maybe worth giving that a try (but without the default [scope](https://github.com/aimad-majdou/main-yarn-files#scope))?

Otherwise, the current workaround is fine by me.

Additional tasks:
- [ ] missing CSS files - comparing output of `gulp log` task (with uncommented paths.styles line):
  ```
  Leaflet.EasyButton/src/easy-button.css',
  leaflet-sidebar-v2/css/leaflet-sidebar.css',
  seiyria-bootstrap-slider/dist/css/bootstrap-slider.css',
  ```
- [ ] update README.md [Build](https://github.com/nrenner/brouter-web#build) instructions

gulpfile.js
```diff
@@ -41,9 +45,12 @@ var paths = {
     'js/control/*.js',
     'js/index.js'
   ]),
-  styles: mainBowerFiles('**/*.css').concat('css/*.css'),
-  images: mainBowerFiles('**/*.+(png|gif|svg)'),
-  fonts: mainBowerFiles('**/font-awesome/fonts/*'),
+  styles: mainNpmFiles().filter(f => 
+    RegExp('.*\.css', 'i').test(f) &&
```
nrenner:
```suggestion
    RegExp('.*\\.css', 'i').test(f) &&
```
As mentioned in my comment, I think the backslash needs also to be escaped.

bagage:
Thanks for the help @nrenner to get it merged.

Indeed one issue with npmFiles is that it ignores `style` property in package.json, so we have to override almost all dependencies and I forgot some. Nice catch!

Apart of that I did the cleaning (removed the `npm install -g gulp` instruction too, because `npm install` will install it locally instead and I think it is cleaner - instead of invoking `gulp` one can use `yarn build` and/or `./node_modules/.bin/gulp` and/or add `./node_modules/.bin` in his/her PATH).

nrenner:
I think we should check in the `yarn.lock` file, what do you think?

[yarn.lock - Check into source control](https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control)
- [Lockfiles should be committed on all projects](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/)

bagage:
Yes indeed we should!

nrenner:
Done.

Thanks!

PR #157, opened 2019-02-18
Render polygons from URL hash and pass it to BRouter server

Phyks:
Hi,

Here is a small increment to my previous work on supporting polygons in BRouter-web.

Polygons are now parsed from URL, rendered and correctly passed to the latest `master` version of BRouter.

screenshot_2019-02-18 brouter web client

You can try this using `BROUTER_WEB_URL/#map=17/48.82463/2.33347/OpenStreetMap&lonlats=2.331714,48.823734|2.336258,48.824221&polygons=2.332106,48.823579,2.33556,48.823579,2.33556,48.824342,2.332106,48.824342,10`.

There are still some things left to do afterwards:

- [ ] The possibility to draw and edit the polygons as well as their weight. I'm not sure I can help a lot on this, I don't know much about the Leaflet plugin currently used :/
- [ ] I'd like to have the possibility to import a GeoJSON file of nogo areas. This is useful to import an OpenData file of roadworks for instance. I think I can work on this if you think this is something worth having.

Best,

### DIFF
--- a/js/plugin/NogoAreas.js
+++ b/js/plugin/NogoAreas.js
@@ -114,7 +114,9 @@ BR.NogoAreas = L.Control.extend({
 
     getOptions: function() {
         return {
-            nogos: this.drawnItems.getLayers()
+            nogos: this.drawnItems.getLayers().filter(function (e) { return e instanceof L.Circle; }),
+            polygons: this.drawnItems.getLayers().filter(function (e) { return e instanceof L.Polygon; }),
+            polylines: this.drawnItems.getLayers().filter(function (e) { return e instanceof L.Polyline; }),
         };
     },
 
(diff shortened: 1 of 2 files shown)


nrenner:
This is causing an error when drawing a nogo circle.

`BRouter.getUrlParams` is expecting the `options` object to be split into nogos (circles), polylines and polygons, but `NogoAreas.getOptions` is returning all in `nogos` from the `drawnItems` FeatureGroup layer:

https://github.com/nrenner/brouter-web/blob/70a0f73d12fde7c4972427bfc475874f9a0ba52c/js/plugin/NogoAreas.js#L115-L119

So I think this should mirror `setOptions` and return 
```
        return {
            nogos: ...,
            polylines: ...,
            polygons: ...,
        };
```

Perhaps by filtering the `drawnItems` FeatureGroup by layer type, e.g. with `instanceof` tests.

Phyks:
Oh, indeed, sorry. I think this should be fixed now.

Phyks:
Thanks!

Phyks:
> Yes, why not.

Ok, this is going to be the next thing I'll work on I think :)

PR #159, opened 2019-02-21
i18n: translate website

### LINKED ISSUE #63: Translations: make website localizable (i18n)

bagage:
Fixes #63. I've used `i18next` + a bunch of plugins:

- `i18next-browser-languagedetector` to detect user language
- `i18next-xhr-backend` to load translations from `.json` files stored on server
- `jquery-i18next` to allow translation for given selector, eg `$('html').localize()`. Maybe we can get rid of this
- `i18next-scanner` to generate .json files automatically from `*.js,index.html`. You can run it via `gulp i18next`.

You can try my branch to test it, you should have French translations if your browser uses this language.

@nrenner please complete the todolist if you see missing items :). Review/feedback are welcome too!

- [x] Add i18n keys for all translatable texts
- [x] Plug with Transifex (pull and push)
- [ ] Should we add a language picker to allow user to pick a different language from his browser preferences?
- [x] translate js leaflet buttons
- [x] translate map layers
- [x] select profile/alternative in navbar

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,18 @@ http://brouter.de
 General BRouter discussions/questions, support:  
 https://groups.google.com/group/osm-android-bikerouting
 
+## Translating
+
+Translations are managed using the
+[Transifex](https://www.transifex.com/openstreetmap/brouter-web/) platform. After
+signing up, you can go to [BRouter's project
+page](https://www.transifex.com/openstreetmap/brouter-web/dashboard/), select a language and
+click **Translate** to start translating.
+
+<script type="text/javascript" src="https://www.google.com/jsapi"></script>
+<script type="text/javascript" src="https://www.transifex.com/_/charts/js/openstreetmap/brouter-web/inc_js/brouter-website/"></script>
+<div id="txchart-brouter-web-brouter-website">Loading chart...</div>
+
 ## Installation
 
 As an alternative to the above online version, the standalone server of BRouter can also be run on your local desktop.
--- a/js/index.js
+++ b/js/index.js
@@ -59,7 +59,7 @@
                     routing.draw(false);
                     control.state('activate-draw');
                 },
-                title: 'Stop drawing route (ESC key)'
+                title: i18next.t('map.draw-route-stop')
             }, {
                 stateName: 'activate-draw',
                 icon: 'fa-pencil',
@@ -67,7 +67,7 @@
                     routing.draw(true);
                     control.state('deactivate-draw');
                 },
-                title: 'Draw route (D key)'
+                title: i18next.t('map.draw-route-start')
             }]
         });
 
@@ -76,11 +76,11 @@
             function () {
                 bootbox.prompt({
                     size: 'small',
-                    title: "Delete route?",
+                    title: i18next.t('map.delete-route'),
                     inputType: 'checkbox',
                     inputOptions: [
                         {
-                            text: '&nbsp;&nbsp;also delete all no-go areas',
+                            text: i18next.t('map.delete-nogo-areas'),
                             value: 'nogo'
                         }
                     ],
@@ -96,7 +96,7 @@
                     }
                 });
             },
-            'Clear route'
+            i18next.t('map.clear-route')
         );
 
         function updateRoute(evt) {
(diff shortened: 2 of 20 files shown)


js/index.js
```diff
@@ -263,7 +276,7 @@
                 var p = {};
                 var sep = '&';
                 if (s.search('&amp;') !== -1)
-                    sep = '&amp;';
+                sep = '&amp;';
```
nrenner: Without the indent this is very misleading now

nrenner:
- I don't like the formatting changes (especially the if statement I commented), were these on purpose?
- Only the two menu items are in French here, will need to check further
- We probably should review/polish all texts before adding keys/translations
  - e.g. I'm unsure about "Plain ascend", "Height diff" is probably more common, but would need to check if it's actually the same thing
- We can add a language picker, once full translation is available
- Do you plan to do the full translation now or just the profile dropdown first?

bagage:
* I'll revert space changes, my editor lit it automatically,no worries here!
* yes I only started with a few items to check if it was OK with you. If so, I'll extract the rest of the keys and complete the French version within this PR.

nrenner:
The missing translations are only in the profile dropdown. This is because bootstrap-select is making a working copy of the `select` tag into a `div.dropdown-menu`, but probably before the translation and without the `data-i18n` attributes.

Otherwise this looks good to me, I think i18next is the right choice, keep up the good work!

bagage:
OK so I've finished @nrenner. I did not do the language picker but I am not sure it would make sense; maybe wait for users feedback first? I did the whole French translations on Transifex to check if it was OK.

If you wish to cleanup strings, do you prefer to do that in this PR or do it later?

bagage:
I've discovered that you can change the language easily by adding `?lng=<new language>` (eg `?lng=de` for instance) in the URL (before the `#` anchor char). Adding a language picker could be easily done as it would only involve modifying this query param.

I also fixed minor issues with langs (thanks to 2 contributors, brouter-web is already available in French, German and partly in Chinese!): added English fallback and handle custom locales (eg `zh_TW`).

README.md
```diff
@@ -16,6 +16,18 @@ http://brouter.de
 General BRouter discussions/questions, support:  
 https://groups.google.com/group/osm-android-bikerouting
 
+## Translating
+
+Translations are managed using the
+[Transifex](https://www.transifex.com/openstreetmap/brouter-web/) platform. After
+signing up, you can go to [iD's project
```
nrenner:
```suggestion
signing up, you can go to [BRouter-Web's project
```

nrenner:
I have no strong opinions about the language picker, we can wait and see if there is a demand for it.

> If you wish to cleanup strings, do you prefer to do that in this PR or do it later?

Not sure anymore if I actually want to do this right now, at least not with this PR.

README.md
bagage: Oops, thanks :)

bagage:
@nrenner I pushed some minor fixes on master directly. I also deployed this version on https://brouter.damsy.net :).  It remains an issue with the `locales` path, eg. XHR requests are done on `brouter.damsy.net/latest/locales` instead of `brouter.damsy.net/latest/dist/locales`. It looks similar to css tricks you did...

nrenner:
I don't know why the previous CSS URL rebasing stopped working with the Yarn setup, so I can't tell if this is similar. The current solution replaces the path part of URLs in CSS files with `images/` (so it's relative to CSS file location in `dist`) and just adds the file name to it. 

The problem this solves is that included projects use different folder structures and thus use different relative paths in their CSS to load images from either same folder, subfolder or parallel folder - but we copy all images to `images` subfolder, so we need to unify the URLs to the images.

Don't know how i18next is interpreting the `loadPath`, relative to the executing JavaScript (dist/brouter-web.js) or relative to web root? Maybe we need to add `dist/...` or `/dist/...`?

bagage:
I've fixed the path problem directly on master (https://github.com/nrenner/brouter-web/commit/1f81c043adaaf1a459ca0fb97ac1bfcaefca96f4), indeed adding `dist/` is enough.

Issue #30, opened 2015-08-07
Load GPX track


nrenner:

poutnikl:
Interesting option would be reusing of already existing **GPX download link** as a "**Tempolink**" for the other user, where it would be on his/her side used as **Load GPX track** . 

He/she could be eventually able to re-download the GPX to store it.

mircozorzo:
A nice to have.

littlehelper:
It would be great if an existing GPX track could be displayed as separate layer, like possible with [Leaflet.FileLayer](https://github.com/makinacorpus/Leaflet.FileLayer).

HankyPanky19:
Hi, it would be great to be able to load an existing GPX file and use that track as starting base for a new routing. Imagine you have developed a long track and finally decide to enclose another new destination - just in the middle of your whole track. Today I must start from scratch and enter each waypoint again and having to include my new destination in the middle. If one could load a track one should be able to shift an exiting path to the new destination with all the necessary recalculation etc.

bagage:
@nrenner since #240 has been merged, what's missing to close this issue?

HankyPanky19:
Hi,
I tried to use the dev version of brouter-web from "brouter.damsy.net" and used feature to load a gpx file. It did display that route, but it seems, there is no way to modify that tour. Maybe I'm missing some steps to accomplish this?!

In my previous comment some months ago I meant, that it would be really great to be able to load an existing gpx file and after load is done treat this track as your currrent work and allow to make changes to that route and finally write back /export that track to a local file again.

Many thanks for your great work!

nrenner:
Loading a *track* is for viewing as a background only.

[GPX](https://en.wikipedia.org/wiki/GPS_Exchange_Format) has two different concepts: track and route. What you want is a route, but that is not supported yet, see #16.

An alternative might be this: the current route is represented in the URL in the Browsers address bar. In order to store a route for later re-use, you can bookmark the current URL in the Browser or copy it as a link into a document.

HankyPanky19:
Well, your suggestion of storing URL bookmarks may help, if you start to design a complex route and you may be able to complete your work at a later time, maybe tomorrow.
What I'm looking for is a mechanism to load a gpx file, which not necessarily was created via brouter and use that "route" as a starting point for changes, e.g. add another destination in the middle of the route. Or use a brouter created gpx track from last years planning and make enhancements to that tour. In these situations there are no URL booksmarks available.
[...]

PR #161, opened 2019-03-03
Let user upload GeoJSON file of nogos

Phyks:
Hi,

This is a very first draft of the ability to create nogos from a GeoJSON file or URL.

screenshot_2019-03-03 client web brouter

2screenshot_2019-03-03 client web brouter

This is definitely not yet ready to get merged, I'm just opening this to get feedback, especially on the UI part.

This can be tested on Paris using this GeoJSON export of impacting roadworks for instance: https://opendata.paris.fr/explore/dataset/chantiers-perturbants/download/?format=geojson&timezone=Europe/Berlin.

TODO:

- [x] read a nogoWeight feature property from the GeoJSON if present and only use the weight input from the dialog as a default if not.
- [x] read nogo circles
- [ ] Add a buffer geometry feature?

Best,

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -20,7 +20,7 @@ body, #content {
 .dataTables_scroll,
 .dataTables_scrollBody {
     display: flex;
-    
+
     flex-flow: column;
 }
 
@@ -61,7 +61,7 @@ table.dataTable {
 .navbar {
     /* align with leaflet-control */
     padding: .5rem 10px;
-}    
+}
 .navbar-brand {
     float: right;
     margin-left: 1rem;
@@ -278,7 +278,7 @@ https://css-tricks.com/svg-line-animation-works/
 }
 .leaflet-bar .fa {
     font-size: 14px;
-    
+
     /* override fa with Leaflet button height for vertical center */
     line-height: 26px;
 }
@@ -340,7 +340,7 @@ table.dataTable.display tbody tr.even:hover {
     background-color: rgba(255,255,0,0.2);
 }
 
-/* 
+/*
  * No-go areas
  */
 
@@ -384,7 +384,7 @@ table.dataTable.display tbody tr.even:hover {
 .leaflet-sidebar-pane#tab_profile,
 .leaflet-sidebar-pane#tab_data,
 .leaflet-sidebar-pane#tab_itinerary {
-    /* Full height for content with inner scrolling, 
+    /* Full height for content with inner scrolling,
        overrides scroll fix for long content in Firefox */
     height: 100%;
 }
@@ -425,9 +425,27 @@ table.dataTable.display tbody tr.even:hover {
     flex: auto;
     /* override default 300px, behaves like min-height with flex auto */
     height: 0;
-    
+
     border: 1px solid #ddd;
-    
+
     font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
     line-height: 1.2em;
+}
+
+#uploadNogos fieldset {
+    display: block;
+    margin-left: 2px;
+    margin-right: 2px;
+    padding-top: 0.35em;
+    padding-bottom: 0.625em;
+    padding-left: 0.75em;
+    padding-right: 0.75em;
+    border: 2px groove;
+}
+
+#uploadNogos legend {
+    display: block;
+    padding-left: 2px;
+    padding-right: 2px;
+    border: none;
 }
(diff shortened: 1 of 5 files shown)


alesarrett:
Just as a feedback, I'm asking whether the input file can accept also line features in addition to areas. The use case might be simply to pre-select some roads that for some reasons are not usable, and upload portions of roads to be used as no-go "areas".
Another clarification is the nogo weight: would it be in your idea a single weight to "calibrate" the no-go?
It would be also great to provide a way to assign different weights to the no go areas/lines in the GeoJSON and to make this understandable by the routing algorithm...

Phyks:
> I'm asking whether the input file can accept also line features in addition to areas. 

It is currently accepting line and polygon features from GeoJSON file. The use case I have in mind is precisely to upload opendata features about road works to avoid them :)

> Another clarification is the nogo weight: would it be in your idea a single weight to "calibrate" the no-go?

This is a penalty used by BRouter under the hood to compute routing. This is basically a cost proportional to the distance travelled inside the nogo area.

> It would be also great to provide a way to assign different weights to the no go areas/lines in the GeoJSON and to make this understandable by the routing algorithm...

This is already possible in the routing algorithm but not in the frontend yet. See #144 for some discussion about it.

nrenner:
>> It would be also great to provide a way to assign different weights to the no go areas/lines in the GeoJSON and to make this understandable by the routing algorithm...

> This is already possible in the routing algorithm but not in the frontend yet. See #144 for some discussion about it.

Although there is no UI for it (yet), the client-side code does support individual weights, you can already pass them with a URL?

So I guess it might make sense to read a `nogoWeight` feature property from the GeoJSON if present and only use the weight input from the dialog as a default if not.

This would probably also enable to add an export option for nogo areas to GeoJSON later and importing them with this feature again (in part related to #16 and abrensch/brouter#91).

Phyks:
I pushed a new commit to improve the handling of loading the nogos from geojson. Nogo weight should be used as a fallback value now and points can be imported as circles.

The main remaining issue is that the URL hash is not updated when loading the geojson file.

nrenner:
> The main remaining issue is that the URL hash is not updated when loading the geojson file.

Updating the route and hash can be done by adding the following calls after `nogos.setOptions`:

```js
    updateRoute({
        options: nogos.getOptions()
    });
    urlHash.onMapMove();
```

> By the way, @nrenner do you know any easy way to buffer a geometry in meters in Brouter-web?

For example with [Turf](https://github.com/Turfjs/turf): http://turfjs.org/docs/#buffer

To work with Leaflet layers I think you would need to convert with [toGeoJSON()](https://leafletjs.com/reference-1.4.0.html#polyline-togeojson) methods and back with [GeoJSON](https://leafletjs.com/reference-1.4.0.html#geojson) layer.

What is your use case?

nrenner:
Some additional thoughts:
- UI
  - As loading additional things is planned (e.g. tracks #30), it might be better to already create a "Load" dropdown menu (like "Download") with "Nogos" as single item for now
  - Not sure, but I personally would prefer "Load" instead of "Upload", as it's not really uploading to and storing on the server, just loading into the browser, what do you think?
  - I would prefer `About` being the last nav item
- Why is the nogo weight mandatory now? Wouldn't it take the default prohibitive value on the server if not set?
- i18n keys are missing (translatable since #159), but we also can do that afterwards
- We haven't used `Promise` and `fetch` yet. I have looked into browserslist tools like [browserslist-generator](https://github.com/wessberg/browserslist-generator) to check if we have visiting Browsers that don't support them and found a few. But I would need to check further and with current logs and will merge as is for now.

Phyks:
> What is your use case?

Oh, great, I missed that Turf could do buffering! Typical use case is when using raw OpenData files for road works for instance, sometimes the lines / polygons are very narrow (in the sense that they cover only the very particular portion of the street which is impacted) but not well positioned (with a slight offset or not covering the whole highway from OSM data). Using a bit of buffering is a very easy solution for this. Another reason is when the polygon for the nogo only cover half of the road, but you expect the road works to actually have impact on a much larger area (traffic jam, etc).

Thanks for all the feedback, I'll have a look at this during the weekend and update the PR!

Phyks:
Hi,

I should have reworked this code a bit to include the remarks (i18n, UX etc) as well as the buffering. Everything is squashed and rebased against the current `master` branch.

The UX part is probably sub-optimal. I think a progress bar could be shown during loading and processing of GeoJSON file for instance, but this is probably fine for a first draft.

This should be ready for review and (hopefully) mergeable. Let me know if you have any question or doubt on this.

Best,

nrenner:
Thanks for the update and your contribution!

> The UX part is probably sub-optimal.

There is always room for improvement, it's fine as it is.

Phyks:
Thanks!

PR #162, opened 2019-03-08
Polish texts for translation

nrenner:
- Add name for vm-forum profiles (see also [Transifex issue](https://www.transifex.com/openstreetmap/brouter-web/translate/#fr/brouter-website/165200325))
- "Trekking bike" instead of just "Trekking" to avoid misinterpretation as hiking
  (as seen in German translation and in the discussion Group)
- consistent use of brackets in profile names

@bagage Are you OK with that? I suppose changes in `en` texts can be just synchronized with transifex without any special consideration?

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -115,20 +115,20 @@
       "custom": "Custom",
       "fastbike": "Fastbike",
       "fastbike-asia-pacific": "Fastbike (Asia Pacific)",
-      "fastbike-lowtraffic": "Fastbike low traffic",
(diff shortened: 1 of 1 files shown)


bagage:
@nrenner Looking good to me. To publish these new strings to Transifex, you can run `yarn transifex-push`. Similarly to fetch new translations from it, use `yarn transifex-pull`.

I can do it if you with though.

nrenner:
> I can do it if you with though.

Yes, please. I guess I need to install Transifex Client first.

Thanks!

PR #175, opened 2019-05-09
Fix for a few errors when loading nogos with weight

### LINKED ISSUE #174: Missing nogo weight when loading nogos
Hi,

It seems loading nogos from a GeoJSON while specifying a (positive) weight does not work and the URL for the query to the BRouter server is not updated accordingly.

It seems that when loading GeoJSON data [here](https://github.com/nrenner/brouter-web/blob/master/js/index.js#L416), the properties of the features in the GeoJSON object are not loaded by Leaflet (they should be loaded into options in Leaflet, for [this part](https://github.com/nrenner/brouter-web/blob/master/js/router/BRouter.js#L338) to work).

I am really not sure what is happening here, I'm almost sure this was fully tested and working when the PR got merged. Maybe this is due to a Leaflet update?

Best,

Phyks:
* Nogo weight was not set when loading nogos with a default weight
value.
* The URL would show up "NaN" values upon reloading the web interface
after having loaded some nogos.

Fix #174.

These fixes are a bit hacky in my opinion, but they work. I'm not sure whether there is a cleaner way to handle this? I did not manage to get much cleaner output from the Leaflet API :/

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -415,9 +415,7 @@
 
                 var geoJSON = L.geoJson(turf.featureCollection(cleanedGeoJSONFeatures), {
                     onEachFeature: function (feature, layer) {
-                        if (!feature.properties.nogoWeight) {
-                            feature.properties.nogoWeight = nogoWeight;
-                        }
+                        layer.options.nogoWeight = feature.properties.nogoWeight || nogoWeight;
                     }
                 });
                 var nogosPoints = geoJSON.getLayers().filter(function (e) {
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -415,8 +415,8 @@
 
                 var geoJSON = L.geoJson(turf.featureCollection(cleanedGeoJSONFeatures), {
                     onEachFeature: function (feature, layer) {
-                        if (!feature.properties.nogoWeight) {
-                            feature.properties.nogoWeight = nogoWeight;
+                        if (!feature.properties.nogoWeight && nogoWeight != -1) {
+                            layer.options.nogoWeight = feature.properties.nogoWeight || nogoWeight;
```
nrenner: This doesn't cover yet the case when `feature.properties.nogoWeight` is set (as it does *not* automatically get copied over, as probably previously assumed here)

Phyks: Indeed, sorry… Thanks for the review! This should be fixed.

PR #176, opened 2019-05-09
Improve about dialog texts

bagage:
Before:

image

After:

image

:warning: when merged we must push to transifex new source file.

### DIFF
--- a/index.html
+++ b/index.html
@@ -137,19 +137,15 @@
                     <p>
                         <i data-i18n="about.contact">Contact:</i><br>
                         <ul>
-                            <li><span data-i18n="about.support">General discussions/questions, support:</span><br>
-                                <a href="https://groups.google.com/group/osm-android-bikerouting" target="_blank">https://groups.google.com/group/osm-android-bikerouting</a>
-                            </li>
+                            <li><a href="https://groups.google.com/group/osm-android-bikerouting" target="_blank"><span data-i18n="about.support">General discussions/questions, support</span></a></li>
                             <li><span data-i18n="about.bug-reports">Bug reports and feature requests:</span>
                                 <ul>
-                                    <li>
-                                        <span data-i18n="about.bug-reports-back">server / backend, routing engine, Android app, profiles, brouter.de site:</span><br>
-                                        <a href="https://github.com/abrensch/brouter/issues" target="_blank">https://github.com/abrensch/brouter/issues</a>
-                                    </li>
-                                    <li>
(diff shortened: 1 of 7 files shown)


nrenner:
That was on a bit on purpose to make the difference explicit, but yes, it's more concise now.

nrenner:
> warning when merged we must push to transifex new source file.

Is that something different than `yarn transifex-push`? Doing a `yarn transifex-pull` after push, now your translations are reverted to english?

Layer translations are deleted in `locales/en.json` and translations (didn't notice that before). Forgot to include `layers/config/overrides.js` in the `i18next` task. Will look at this tomorrow.

nrenner:
Ok, my translations mess should be fixed now, this is what I have done:
- restored layer keys (d8886e7) and texts (66cdc84)
- `tx push --source` to add the restored layer keys
- `tx push -t -f` to force push the translations of this PR and to restore the deleted layer translations
- restore changes overwritten by force (manually on Transifex) since last update, only one found:
  - fr: Ascend (Plain ascend) = Dénivelé cumulé (dénivelé global)
- `yarn transifex-pull`

nrenner:
Just got a message from Transifex that all languages are now 100% translated. Turns out force pushing language files wasn't such a good idea, as all previously untranslated texts are now set to their `en` fallback text.

Any idea how to reset the `en` texts, except doing it manually?

At least there is a time-based filter, e.g. [`translated_after:2019-05-05`](https://www.transifex.com/openstreetmap/brouter-web/viewstrings/#zh_TW/brouter-website/165200255?q=translated_after%3A2019-05-05).

bagage:
Hm you are right. Sorry about that mess. 

Just to be clear, but I believe you understood it by now: `yarn transifex-push` will only push English version to transifex, not other languages. Since it was only some typography stuff I wanted to do it directly but I should have done it on Transifex instead.

Anyway, I'll take a look at that and let you know if I can find something.

bagage:
I don't see any option on transifex site directly. However something that may work would be to force push  `json` files after having removed translations that are identical to English, for all files. Only non translatable / identic words would be lost, but there are not many I believe.

I can do it if you wish @nrenner.

nrenner:
Thanks for helping to look for a solution.

I thought about it and I think deleting the english texts manually on Transifex using above filter is probably the best and safest option. It's only about `hu`, `pt` and `zh_TW` anyway.

I can do it probably tomorrow afternoon, but you can also do it if you feel like it, I don't mind either way.

bagage:
Should be fixed :).

nrenner:
Great, thanks!

PR #178, opened 2019-05-09
Set trackname in downloaded file and set custom name

### LINKED ISSUE #96: File and Track title when exporting
Hello,
when exporting the created track I'd like to define the filename and track name to have a better import of it on the other devices.
Thank you for your this useful service

bagage:
Fixes #96.

I thought a bit about that and I think:

- a default **localized** name should be generated, by default:
  - if the start and end is within the same city, then `loop around {city_start} - {distance}km`
  - otherwise `from {city_start} to {city_end} - {distance}km`
- if route is modified, the name is regenerated
- Ideally this name should be editable on the website - if edited, then we won't regenerate it again.

WDYT?

Left to do:

- [x] do not query geocoding multiple times (sometime there are 10 requests sent simultaneously)
- [x] improve name CSS/design (move out of navbar?)
- [x] localize name
- [x] use async xhr request
- [x] wait upstream merge https://github.com/abrensch/brouter/pull/157 https://github.com/abrensch/brouter/pull/158


nrenner:
I'm concerned that this is sending way too many requests when automatically updating on every route change. And this is for nothing really, as the name is only needed when actually downloading the final route.

My suggestion would be to turn the "Download" dropdown into a modal dialog, add the name input field there and offer a button to auto-fill the name.

bagage:
Indeed it's much simpler via a new modal - apart from last brouter-engine fix, I think I am done with this.

Thanks for reviewing @nrenner :).

nrenner:
This example doesn't return names:
http://brouter.de/brouter-web/#map=12/47.6720/9.3602/standard&lonlats=9.479958,47.649113|9.270487,47.692577

I would expect `town` (but not `suburb`) to be read here instead of `village`, and perhaps also considering more place values:
https://wiki.openstreetmap.org/wiki/Key:place#Populated_settlements.2C_urban

bagage:
Thanks for the wiki link. I modified the name to return `addr.village || addr.town || addr.hamlet || addr.city_district || addr.city`. Does it sound good to you?

nrenner:
Don't really know what values make sense here, but looks good to me.

bagage:
Upstream fixes were deployed so I believe this is good to go :).

:warning: since there are new/replaced strings, after merged we have to run `yarn transifex-push`.

PR #179, opened 2019-05-11
Add reverse route / delete last point buttons

### LINKED ISSUE #33: Delete last point
Hello
this is more a feature request: like in GPSies possible, in the brouter-web there is strongly a button missing to delete just the last point (or I am not finding it).
It is possible to delete the whole track, and it is possible to move the last point, but not possible to delete the last point.
Possible to implement?
Thanks

frank

### LINKED ISSUE #54: button for reversing the direction of the track
Hi!
One feature I am missing quite a lot is, reversing the track.
if you have the following points: Start - A-B-C-D-Finish, clicking the "new" button should trigger the generation of a track: Finish-D-C-B-A-Start

Is is possible to implement?
Thanks Bosdla

bagage:
Fixes #54
Fixes #33

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -35,6 +35,7 @@
             deleteButton,
             drawToolbar,
             urlHash,
+            reverseRoute,
             saveWarningShown = false;
 
         // By default bootstrap-select use glyphicons
@@ -69,6 +70,14 @@
                 title: i18next.t('map.draw-route-start')
             }]
         });
+
+        reverseRouteButton = L.easyButton(
+            'fa-random',
+            function () {
+                routing.reverse();
+            },
+            i18next.t('map.reverse-route')
+        );
 
         deleteButton = L.easyButton(
             'fa-trash-o',
@@ -227,7 +236,7 @@
 
         routing.addTo(map);
         elevation.addBelow(map);
-
+        
         sidebar = BR.sidebar({
             defaultTabId: BR.conf.transit ? 'tab_itinerary' : 'tab_profile',
             listeningTabs: {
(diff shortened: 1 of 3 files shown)


nrenner:
No sure we should use the undo icon (fa-undo) for something that isn't really an undo function, maybe one day we get around actually implementing a proper undo feature.

Maybe something like [fa-long-arrow-left](https://fontawesome.com/v4.7.0/icon/long-arrow-left) [stacked](https://fontawesome.com/v4.7.0/examples/#stacked) with [fa-map-marker ](https://fontawesome.com/v4.7.0/icon/map-marker) instead?

bagage:
I started to replace undo icon but currently it does not look OK:

image

I'm not sure how we can stack then vertically, if you have an idea feel free to modify this MR directly :).

nrenner:
It gets horizontally aligned without `fa-lg`, and I moved the marker with margin and colored it gray so they're better distinguishable, but I still don't like it:
Screenshot_2019-05-15 BRouter Web Client(2)
Screenshot_2019-05-15 BRouter Web Client(3)

```html
'<span class="fa-stack"><i class="fa fa-map-marker fa-stack-1x" style="margin: -2px 0 0 2px; color: gray;"></i><i class="fa fa-long-arrow-left fa-stack-1x"></i></span>',
```

I guess this one is better? Marker gray to indicate fading and to highlight left caret as back/backspace.

Screenshot_2019-05-15 BRouter Web Client(1)

```html
'<span><i class="fa fa-caret-left"></i><i class="fa fa-map-marker" style="margin-left: 1px; color: gray;"></i></span>',
```

bagage:
Yes last one is looking good IMHO.

PR #182, opened 2019-05-16
Add prettier and reformat code

bagage:
This formats code automatically before `git commit` automatically, using git `pre-commit` hook.

`yarn prettier` will force reformatting of all project, but it should do nothing in the future, as soon as we use git hooks.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,9 @@
+LICENSE
+dist/
+*.zip
+yarn.lock
+.idea
+.gitignore
+.prettierignore
+.tx/
+layers/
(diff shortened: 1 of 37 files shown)


nrenner:
Thanks, that was also on my Todo list, as well as ESLint.

I immediately dislike just two spaces and double quotes, but I think we should stick with the defaults.

What I noticed:
- `yarn.lock` is in the ignore list, yet still reformatted?
- All files in subfolders of `js` are not reformatted?

bagage:
I'm not fan of 2-spaces too, I added a config to tweak these. I agree with you that we should stick as much as possible to default values, but this is literally [the example they give in their own doc](https://prettier.io/docs/en/configuration.html) - I believe it's a strong "dispute" across projects.

PR #185, opened 2019-05-19
Ensure that nogo weight is not null nor undefined before using it

### LINKED ISSUE #183: nogo's are broken in current release
without a nogo, the server sees:

GET /brouter?lonlats=8.61886,49.692536|8.615341,49.699832&profile=trekking&alternativeidx=0&format=geojson

with a nogo, this gets:

GET /brouter?lonlats=8.61886,49.692536|8.615341,49.699832&nogos=8.614697,49.696536,105,undefined&profile=trekking&alternativeidx=0&format=geojson HTTP/1.1

and that results in a parsing error.

this is with or without the latest pull request on the brouter-repo applied, so looks like something is missing for parsing nogo-weights ?

bagage:
Fixes #183

### DIFF
--- a/js/router/BRouter.js
+++ b/js/router/BRouter.js
@@ -40,7 +40,6 @@ L.BRouter = L.Class.extend({
 
     getUrlParams: function(latLngs, format) {
         params = {};
-
         if (this._getLonLatsString(latLngs) != null)
             params.lonlats = this._getLonLatsString(latLngs);
 
@@ -269,6 +268,7 @@ L.BRouter = L.Class.extend({
             // -1 is default nogo exclusion, it should not be passed as a URL parameter.
             if (
                 circle.options.nogoWeight !== null &&
+                circle.options.nogoWeight !== undefined &&
                 circle.options.nogoWeight !== -1
             ) {
                 s += L.BRouter.NUMBER_SEPARATOR;


js/router/BRouter.js
```diff
@@ -267,10 +266,7 @@ L.BRouter = L.Class.extend({
             s += L.BRouter.NUMBER_SEPARATOR;
             s += Math.round(circle.getRadius());
             // -1 is default nogo exclusion, it should not be passed as a URL parameter.
-            if (
-                circle.options.nogoWeight !== null &&
-                circle.options.nogoWeight !== -1
-            ) {
+            if (circle.options.nogoWeight && circle.options.nogoWeight !== -1) {
```
nrenner: Testing falsy values is also excluding `0`, which I'm not sure is what we want. But I actually don't know what the sensible/valid value range for `nogoWeight` is. @Phyks?

Phyks: `nogoWeight` is read from the URL in the BRouter server side at https://github.com/abrensch/brouter/blob/master/brouter-server/src/main/java/btools/server/request/ServerHandler.java#L285-L289.

Then, it is used in BRouter as:
* If it is `-1` or `NaN`, it is an excluding nogo (similar to the previous mechanism, completely avoid the area).
* If it is `>= 0`, this is a weight for the nogo area.

I'm not sure about the issue here though, as testing `!== null` should match `undefined` as well, isn't it?

I have 

```
> undefined !== null
true
```

in my browser console.



Anyways, a nogo area with a zero weight is simply useless. It adds more processing time on the server side (the cost for travelling through the nogo is still computed, but multiplied by zero in the end :/).

bagage: @Phyks I think we agree. `null` and `undefined` are different, so `undefined !== null` returns `true` as expected.

The issue here is that `undefined` should be treated as `null`?

Phyks: Indeed, sorry I messed up in my latest message. `undefined` should be treated as `null` indeed. Thanks!

nrenner: I would intuitively use 0 to disable a no-go (for whatever reason).

Not sure if the constraints for the costfactor also apply to the no-go weight:

> - The costfactor is required to be >= 1, otherwise the cost-cutoff
  logic of the routing algorithm does not work and you get wrong results.

http://brouter.de/brouter/profile_developers_guide.txt

But validating for a useful value range would be a different issue and propably done at a different place.

PR #186, opened 2019-05-19
Bootstrap upgrade

### LINKED ISSUE #107: Bootstrap 4.0 update and migration
https://getbootstrap.com/docs/4.0/migration/

- [x] last version supporting Bower: v4.0.0-beta, currently using: 4.0.0-alpha.5
  `bower install bootstrap#v4.0.0-beta`
  [Drop Bower support (#23568) · twbs/bootstrap@c770e68](https://github.com/twbs/bootstrap/commit/c770e6883695e8b5d92a1107d3a49b1cb63655b6) **we do not use bower anymore**
- [ ] [Responsive utilities](https://getbootstrap.com/docs/4.0/migration/#responsive-utilities) removed
  - see also current alpha docs: https://v4-alpha.getbootstrap.com/layout/responsive-utilities/
  - [ ] `.hidden` class -> `hidden` attribute (IE 11-, with Bootstrap _normalize.scss/[_reboot.scss](https://github.com/twbs/bootstrap/blob/27f553c9651dbe81a8aaf821aa2bf15670d9adb8/scss/_reboot.scss#L478-L482): IE 10-)
  - [ ] `hidden-sm-down` (index.html), ...
- [ ] use [flexbox utilities](https://getbootstrap.com/docs/4.0/utilities/flex/) instead of own classes?
- [ ] ...

bagage:
Fixes #107 

I'm not sure how this should be tackled. For now I'm just trying to fix broken things. 

Still to do:

- [x] reput the brand in navbar right corner? What was the purpose of it? cf https://stackoverflow.com/questions/49434504/using-bootstrap-4-how-to-align-navbar-brand-on-the-right
- [x] image
- [x] image
- [x] image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -65,11 +65,6 @@ table.dataTable {
     /* align with leaflet-control */
     padding: 0.5rem 10px;
 }
-.navbar-brand {
-    float: right;
-    margin-left: 1rem;
-    margin-right: 0;
-}
 .nav-link .fa {
     margin-right: 0.2em;
 }
@@ -262,7 +257,7 @@ https://css-tricks.com/svg-line-animation-works/
     background: none;
     outline: none;
 
-    /* bootstrap .btn-default */
+    /* bootstrap .btn-secondary */
     background-image: linear-gradient(to bottom, #fff 0, #e0e0e0 100%);
     background-repeat: repeat-x;
     box-shadow: inset 0 1px 0 rgba(255, 255, 255, 0.15),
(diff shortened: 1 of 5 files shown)


nrenner:
> I'm not sure how this should be tackled. 

Don't know either, the [migration guide](https://getbootstrap.com/docs/4.0/migration/) is a bit long and we don't really use that much. Maybe by going through `index.html` and checking everything that looks like Bootstrap?

> reput the brand in navbar right corner? What was the purpose of it?

Yes, please. To have the profile dropdown in the leftmost position of the screen, in order not to hide the route while trying out various profiles and see how the route changes. Also, I thought it just looks weird to open this huge dropdown somewhere in the middle of the screen.

bagage:
OK I think this is it - reviews are always welcome :).

bagage:
Thanks, indeed I missed some. Fortunately there were easy fixes.

Regarding the modal issue, I have to upgrade `bootbox` dependency but gulp is not finding it, eg `gulp log` is not listing `gulp-debug: node_modules/bootbox/dist/bootbox.all.min.js` as expected (even when I manually override `main` entry). I'm trying to understand why but yet I failed to do so…

bagage:
Hmm, we are actually filtering out `*.min.js` files in gulp process. Is that on purpose? 

One way is to use the non minified file instead:

```json
        "bootbox": {
            "main": [
                "src/bootbox.all.js"
            ]
        },
```

But the other way would be to just use `.min.js` files as well. I know that we minify file later in the process for releases, but do we need non-minified libraries too?

nrenner:
Thanks a lot for taking on this tedious task!

nrenner:
> Hmm, we are actually filtering out *.min.js files in gulp process. Is that on purpose?

I think so, but can't remember why. Probably because some projects had both the src and min files in their `main`.

I think I haven't seen a case yet where we would have needed the minified file, non-minified or source files were always available and sufficient.

[bower.json spec](https://github.com/bower/spec/blob/master/json.md#main) says:
> Do not include minified files.

The reason for me is to have non-minified library code available for debugging and proper stack traces, both in debugging mode and also in the release sourcemap.

PR #189, opened 2019-05-21
Add link to #brouter@OFTC

bagage:


README.md
```diff
@@ -70,7 +75,7 @@ As an alternative to the above online version, the standalone server of BRouter
 
 ### Run
 
-1.  start BRouter server in the `standalone` directory with `./server.sh` or `server.cmd` (Windows)
+1.  start BRouter server with `./misc/scripts/standalone/server.sh` or `server.cmd` (Windows)
```
nrenner: This refers to the "Install BRouter" section above, which is based on the release zip, not the source repository. So the previous description is still valid (the recent script merge is not released yet).

bagage: Indeed, my bad :).

PR #199, opened 2019-06-01
Fix typo: opencyle -> opencycle

rkflx:


nrenner:
Thanks, but that's the id of the original JOSM definition:
https://josm.openstreetmap.de/wiki/Maps/Worldwide#OpenCycleMap

So we should keep it, otherwise updating fetches wouldn't match that entry anymore and I wouldn't want to change the id somewhere inbetween.

PR #200, opened 2019-06-01
Polish formatting and behaviour of track statistics bar

rkflx:
This tries to make the overall experience of using the track statistics bar a bit smoother. In particular:

Formatting:
- Use "x | y" format instead "x (y)" (The old format was a bit confusing, e.g. the second part of the label could be mistaken for an "explanation" of the first part, while it actually is an entirely separate label.)
- Repeat unit for every metric.
- Use plural for units.
- Slightly change wording ("Energy per 100km").
- Use locale-aware number formatting (i.e. add grouping separators).
- Use hours and minutes to format travel time.

Behaviour:
- Use "-" everywhere when a metric is not yet available instead of mixing "-" and "0".
- Properly revert to initial "-" when removing route instead of showing "0".
- Show "0" instead of "" when route length cannot be calculated yet.

Localization:
- Separate formatting from translations

### DIFF
--- a/index.html
+++ b/index.html
@@ -952,11 +952,14 @@
                         </p>
                     </li>
                     <li hidden>
-                        <div
-                            class="text-muted small hidden-sm-down"
-                            data-i18n="footer.total-energy"
-                        >
-                            Total Energy | Energy per 100 km
+                        <div class="text-muted small hidden-sm-down">
+                            <span data-i18n="footer.total-energy"
+                                >Total Energy</span
+                            >
+                            |
+                            <span data-i18n="footer.energy-per-100km"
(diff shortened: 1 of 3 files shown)


nrenner:
Nice improvements, thanks for your contributions!

rkflx:
Thanks for merging!

There's one more patch I have related to the statistics bar: "Always show travel time and energy, so the layout does not move around so much when starting to draw a route." However, I guess we'll have to wait for an official BRouter release containing that feature first. I'll (try to) keep you posted.

rkflx:
> There's one more patch I have related to the statistics bar: [...] I'll (try to) keep you posted.

Done in #216.

PR #208, opened 2019-06-16
Add chat in about modal

bagage:
:warning: new string, hence `make transifex-push` must be run after merge (I can do it if needed)

### DIFF
--- a/index.html
+++ b/index.html
@@ -337,6 +337,15 @@
                                 >
                             </li>
                             <li>
+                                <a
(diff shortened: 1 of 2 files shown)


nrenner:
Done the push and also a pull, not sure if it always makes sense to do a pull after push?

bagage:
I dont think its necessary indeed since in case of a missing translation it will use the fallback version from html directly.

PR #209, opened 2019-06-20
Bump gulp to version 4.0.2

Phyks:
Hi,

Having `node` version `v12.4.0` on my laptop, I was not able to build `brouter-web`. It seems it is enough to bump Gulp to version 4, which is done in this PR.

Best,

P.S.: Btw, it would be super useful to have a watch mode (there is currently only a `yarn run build` command) with auto-rebuild on file change. Not sure if `gulp` can do it :/


nrenner:
Thanks.

You probably mean `yarn build watch` (see [readme](https://github.com/nrenner/brouter-web#develop))?

Phyks:
Thanks!

> You probably mean yarn build watch (see readme)?

Indeed! Sorry, I completely missed it. What about adding a `yarn watch` alias?

nrenner:
I don't mind adding an alias.

But now the watch task doesn't work anymore for me, see #210.

Issue #217, opened 2019-07-16
Delete nogo areas


Phyks:
Hi,

There is currently no way in BRouter-web to delete only the nogo areas, which could be especially useful with the new "upload" feature.

What about switching the trash icon to simply "Delete" and in the next modal having two checkboxes, (eventually ticked)?

- [x] Delete the itinerary
- [ ] Delete the nogos

Best,

nrenner:
Yes, would make sense.

Not sure about the wording, maybe:
- [x] Delete route
- [ ] Delete all nogos

PR #224, opened 2019-08-04
Let user delete route and/or nogo areas

Phyks:
Here is a proposal for a fix for #217. 

How should translations be handled? Should I update them here along the code (at least some of them)? Or is everything based on a third-party tool?

Best,


nrenner:
I'm unsure about the best practice with translations myself, but I guess it would nice to have the english language file updated along with a pull request.

The keys can be extracted/synched from the code with `yarn build i18next`. The english original texts then need to be added manually to `locales/en.json` (even when already provided in the code?).

The translations themselves are done on the Transifex platform (editing them manually leads to conflicts). The upload of new keys (en.json) is usually done by me after merging a PR (unlike with the recent ones).

Phyks:
Ok! sorry for coming back late on this, I should have updated the translations as well.

Issue #129, opened 2018-08-17
Use distinct from, to and via markers


bagage:
It seems there is no way today to know in which the route is going:

![image](https://user-images.githubusercontent.com/1451988/44277359-c0291e80-a24a-11e8-8d9f-e0c75e2f53ad.png)

It would be nice to have different icon and/or color for start marker and end marker, for instance on osm.org:

![image](https://user-images.githubusercontent.com/1451988/44277745-e4d1c600-a24b-11e8-91c2-3434ce0157c3.png)

nrenner:
I agree, it's just not implemented yet (Turistforeningen/leaflet-routing#8).

An idea was to first evaluate whether it would make sense to switch to [leaflet-routing-machine](https://github.com/perliedman/leaflet-routing-machine/), which is lacking other features.

bagage:
Thanks for the quick insight @nrenner. The main advantage of leaflet-routing-machine I see is that it is still under development, but I do not know much more about each project and the features its lacking.

If it requires huge work to be replaced, maybe I'll give leaflet routing icons configuration a try.

nrenner:
I updated the `dev` branch of my leaflet-routing fork that we reference with your PR. You should be able to update to that with `yarn upgrade leaflet-routing`.

Issue #68, opened 2017-02-26
Allow straight lines


bagage:
Would it be possible to draw some straight lines? It is mostly useful when you know that a route is there but not present on OSM right know; instead of making a detour you just want to have a straight line. 
I don't know how much effort it would require but it would be definitely help, ie I mainly would use it to map actually missing ways.

- [x] client-side track formatting (#399)
- [x] extend [leaflet-routing](https://github.com/nrenner/leaflet-routing/commits/dev) plugin
- [x] support in all components (hash, distance markers, ...) (#497)
- [x] UI -> follow-up issue #498
- [x] some remaining tasks

*<sub>edit (nrenner): task list<sub>*

nrenner:
Examples are gpsies.com and outdooractive.com, see also Discussion 
[BRouter, routing problem, manuell selection way (bridge /ferry) - manuelle Routenauswahl für Teilabschnitt (Brücke / Fähre)–Google Groups](https://groups.google.com/forum/#!topic/osm-android-bikerouting/KeyMl01s1Bo) (German)

lorissaa:
Hello, i search for a same function!
gpsies.com has such a function: in the sidebar with a checkbox "follow roads". Is this by routing always checked, follow ways, and unchecked on during routing makes direkt lines, and continue routing on ways checked the checkbox.
Is it possible to install it in brouter-web?

gimoya:
IMHO this is a very important feature! Please consider an implemantation, if possible! Thanks!

nrenner:
Some more requests in the Group:
- [BRouter Web - waypoints and shortcuts](https://groups.google.com/d/topic/osm-android-bikerouting/q0lW6Ib3hV4/discussion)
- [Is any way to add no-routed line to route?](https://groups.google.com/d/topic/osm-android-bikerouting/icJF84SsxNw/discussion)
- [Changing Profile while Routing](https://groups.google.com/d/topic/osm-android-bikerouting/2YHDLc4gJNI/discussion)

Probably related server issue: abrensch/brouter#145. 

Seems like the most popular feature request right now, adding it to the "next" milestone.

[...]

PR #226, opened 2019-08-06
export user's POI (point of interest)

### LINKED ISSUE #222: Add POIs in the exported route
I thought that there was already an issue about that, but I'd like to:

- draw points on BRouter website, and enter a name for each
- then, when exporting the route (at least as GPX), export these points as `waypoints`. 

GPX syntax is:

```xml
<gpx>
  <wpt lon="3.866962" lat="43.674468">
    <name>My POI to visit</name>
  </wpt>
 <trk>...</trk>
</gpx>
```

bagage:
Fixes #222 

Still left to do:

- [x] extract translations
- [x] input validation: how to ensure that POI name is valid? For gpx there is no limitation (except XML itself), not sure about the others.
- [x] ~~allow user to chose icon type~~? [cf Leaflet.awesome-markers](https://github.com/lvoogdt/Leaflet.awesome-markers) **for now, all POI are displayed with :star: icon. We can extend it quite easily if needed** 
- [x] persist markers in URL (from / to)
- [x] refresh url on marker modification (move/delete/add)
- [x] write markers in generated file, either by passing them to brouter ~~, or by editing the file after it has been generated~~
- [x] custom marker icon when drawing POIs?
- [x] escape should cancel drawing mode as route?
- [x] drawing route and POI edition should be exclusive
- [x] shortcut to start drawing POI
- [x] allow to remove all POI in delete button menu
- [x] allow to remove POI from popup

Questions:

- [x] does it make sense to add the POIs layer in layers control so that it can be shown/hidden? Or should it be added to the map directly


bagage:
@nrenner what about dropping all file formats and only use one (say, gpx), then external libraries to convert from GPX to anything else?

It would ease developing new features related to file - yet I'm not sure how to tackle it otherwise.

Phyks:
My two cents on this (did not yet have time to test the code locally), while you are changing the POIs and the markers, what about adding a distinction between markers (start / end and waypoints)?

bagage:
Would be great @Phyks, but it is actually another issue yet-non-trivial: #129.

nrenner:
> what about dropping all file formats and only use one (say, gpx), then external libraries to convert from GPX to anything else?

What do you mean exactly with "dropping all file formats" and converting? Ignore the other `formatAs*` methods in BRouter [OsmTrack](https://github.com/abrensch/brouter/blob/6cc7f07f4de2f47311e10c58dece3e5368c1e681/brouter-core/src/main/java/btools/router/OsmTrack.java#L400) and convert in the client? But still pass all values (user POIs here) to the server to get the primary format (GPX?)? With the `download` attribute, we currently don't get hold of the server result, that would need to be changed.

bagage:
> What do you mean exactly with "dropping all file formats" and converting? Ignore the other `formatAs*` methods in BRouter [OsmTrack](https://github.com/abrensch/brouter/blob/6cc7f07f4de2f47311e10c58dece3e5368c1e681/brouter-core/src/main/java/btools/router/OsmTrack.java#L400) and convert in the client? 

Yes.

> But still pass all values (user POIs here) to the server to get the primary format (GPX?)? 

No, in this case we would insert them directly in the client, after we got the gpx generated. The alternative is to modify backend to accept a list of POIs and write them for each file format.

> With the `download` attribute, we currently don't get hold of the server result, that would need to be changed.

In that case, yes.

Not sure which solution is better yet.

nrenner:
Not sure either. After the discussion in abrensch/brouter#91 and with trackName already implemented it seemed more straightforward and reasonable to do those in the server, so I sort of gave up the idea of client-side export formatting.

And with route waypoint export already implemented in the server, I'm not sure it's worth starting a different way just for user POI waypoints. 

Future features regarding export that come to mind:
- #16 (GPX route download + upload): with route waypoints now implemented in the server it probably makes sense to do the same with `rte`, and might be useful for other clients too
- #196 (Track Timestamp?): time deltas currently not available, would need to be added in the server anyway
- #68 (Allow straight lines): converting from server GPX won't help here; to do this client-side we would probably need to completely replace the current export that sends all route points to the server,  aggregate the route segments we already have ourselves and probably convert from GeoJSON

@abrensch would you mind user POI waypoints getting passed to OsmTrack just for file formatting or would you prefer those handled completely in the client?

Phyks:
> No, in this case we would insert them directly in the client, after we got the gpx generated.

Another advantage of this (thinking of https://github.com/nrenner/brouter-web/issues/231#event-2564039136) is that BRouter-web could add a comment in the GPX file with the URL used to generate the track. Not sure this is worth the pain though.

bagage:
For now, and based on @nrenner, I'll probably go for the first solution (implement POI export in backend) - URL could be another query param option (`?description=Generated with URL <url>` or so)

nrenner:
Not sure, but I think the server request URL is pretty much the same as in the client, so a simple regex replacement would probably do - or at least the server gets all information to reconstruct the important part of the client URL (except for the `map` parameter).

There is an issue for this: abrensch/brouter#152.

bagage:
OK I'm done with it. brouter counterpart is https://github.com/abrensch/brouter/pull/196.

Comments are welcome :).

bagage:
Basically looking like:

image

js/plugin/POIMarkers.js
```diff
@@ -0,0 +1,138 @@
...
+            }
+        });
+    },
+
+    addMarker: function(latlng, name) {
+        var icon = L.AwesomeMarkers.icon({
+            icon: 'star',
+            markerColor: 'cadetblue'
+        });
+        var content = name + '<br>';
```
nrenner: The user input `name` should be sanitized/encoded here to prevent [Cross-site Scripting (XSS)](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) (DOM based).

For example this URL injects a JavaScript alert that is executed when the popup opens:
http://localhost:3000/#map=6/47.636/7.086/standard&pois=7.58045,47.576526,%3Cimg%20src%3D%22xyz%22%20onerror%3D%22alert('script%20executed')%22%3E

The most simple fix usually is to use `Node.textContent` instead of `innerHTML`, but in this case it is used by Leaflet internally in [DivOverlay.js](https://github.com/Leaflet/Leaflet/blob/56cdbf69a91465e1e1db047e17cc388e53a74652/src/layer/DivOverlay.js#L201). We could also use a lib like [DOMPurify](https://github.com/cure53/DOMPurify).

js/router/BRouter.js
```diff
@@ -231,6 +236,39 @@ L.BRouter = L.Class.extend({
         return lonlats;
     },
 
+    _getLonLatsNameString: function(latLngNames) {
+        var s = '';
+        for (var i = 0; i < latLngNames.length; i++) {
+            s += this._formatLatLng(latLngNames[i].latlng);
+            s += L.BRouter.NUMBER_SEPARATOR;
+            s += latLngNames[i].name;
```
nrenner: Then `name` part should be encoded with `encodeURIComponent`.

Otherwise e.g. a name like `a&b` will break the parsing later (not showing the second POI and only "a" as name): 
http://localhost:3000/#map=6/50.965/8.053/standard&pois=2.351632,48.850258,a&b;13.338911,52.536273,c

We are not encoding the whole URL for better readability and because we know it is not needed, but we cannot control user input, therefore such parts do need encoding.

js/router/BRouter.js
```diff
@@ -231,6 +236,39 @@ L.BRouter = L.Class.extend({
...
+
+        if (!s) {
+            return lonlatnames;
+        }
+
+        groups = s.split(L.BRouter.GROUP_SEPARATOR);
+        for (var i = 0; i < groups.length; i++) {
+            // lng,lat,name
+            part = groups[i].split(L.BRouter.NUMBER_SEPARATOR);
+            lonlatnames.push({ latlng: L.latLng(part[1], part[0]), name: decodeURI(part[2]) });
```
nrenner: I think for just decoding parts of a URL this should be `decodeURIComponent` instead of `decodeURI`.

bagage: indeed, my bad :).

js/plugin/POIMarkers.js
bagage: Regarding [this article](https://gomakethings.com/preventing-cross-site-scripting-attacks-when-using-innerhtml-in-vanilla-javascript/), I added the provided sanitizer function: 

```js
var sanitizeHTML = function (str) {
            var temp = document.createElement('div');
            temp.textContent = str;
            return temp.innerHTML;
        };
```
WDYT?

nrenner: Will probably work for our use case, but not when used for attribute values, according to that old article:
[You are probably misusing DOM text methods](https://web.archive.org/web/20121208091505/http://benv.ca/2012/10/4/you-are-probably-misusing-DOM-text-methods/)

So maybe rename to `sanitizeHTMLContent` and add a comment to not use for attributes.

bagage: Since the method is internal to POI markers, I believe we won't use it for something else but indeed I added a comment about that. Thanks!

nrenner:
Thanks!

nrenner:
I moved the POI button below the routing toolbar, because I found it too prominent for a secondary feature and better in line this way with the probable order of user actions. Let me know if you disagree, I'm open for discussion.

bagage:
No problem @nrenner, I think it makes sense!

PR #227, opened 2019-08-07
Nogo fixes for GeometryCollection geojson + minor UI

bagage:
I had issue with this geojson, so I had to use `turf.flattenEach` instead of `turf.featureEach` - the only difference seems to be that flattenEach handles GeometryCollection while featureEach does not.


bagage:
@Phyks glad if you could review it :).

Phyks:
Looks good to me. Thanks for the fixes!

Issue #220, opened 2019-07-16
NS_ERROR_MALFORMED_URI when importing nogos with buffering


Phyks:
Hi,

When importing a GeoJSON file of nogo areas (typically https://opendata.paris.fr/explore/dataset/chantiers-perturbants/download/?format=geojson&timezone=Europe/Berlin) with some buffering, this results in a `NS_ERROR_MALFORMED_URI` being emitted (`brouter-web.js:93`, seems to be happening in `onMapMove leaflet-fullHash.js:192`).

Then, the modal would not close and the URL would not be updated with the correct parameter.

The buffering action works fine though and the nogos are displayed on the map. The route computation fails, due to a subsequent `TypeError: t is null` in `L.Routing.Draw`.

Phyks:
I had a look at it and my best bet so far is that it is not particularly due to buffering. The issue is probably that all polygons and nogo areas get appended to the URL, which makes it too long at some point and considered as being malformed.

Not sure what can / should be done then.

nrenner:
I can reproduce in Firefox, works fine in Chromium.

My guess would also be that the URL gets too long, but before taking measures, I think we would need to be sure, e.g. by researching Firefox URL limit; narrow down the error by splitting the polygons parameter of the URL into halves and see if both are valid, ... Would leave that to you unless you get stuck as I really want to get on with [milestone 0.11.0](https://github.com/nrenner/brouter-web/milestone/7).

Not sure if that would really help, my first thought is to try same kind of polyline encoding, like [mapbox/polyline](https://github.com/mapbox/polyline) (we already include that one).

Phyks:
I tested with a very long GeoJSON, trimming it to only keep part of the features, and I am almost sure this comes from the URL length (this is happening when too many features are in the GeoJSON and this is not bound to a specific feature).

I tried to run 

```js
url = 'http://example.com#'; for (i = 0; i < 2000000; ++i) { url = url + 'a'; }; new URL(url)
```

in a console in Firefox. This crashes with a `TypeError` (the limit is between 1e6 and 2e6) but this works in Chromium.

Indeed, we should probably have a compact encoding of the polylines / polygons. `mapbox/polyline` is probably a solution. I'll have a look once #227 is fine and merged to avoid conflicts.

PR #233, opened 2019-08-19
Use mapbox/polyline for polyline/polygon URI param encoding

Phyks:
Hi,

Looking at https://github.com/nrenner/brouter-web/issues/220, it seems that `mapbox/polyline` encoding is indeed enough to get past the URI length issue. Here is a first draft, not yet functional. :/

### DIFF
--- a/js/router/BRouter.js
+++ b/js/router/BRouter.js
@@ -283,22 +283,16 @@ L.BRouter = L.Class.extend({
     _getNogosPolylinesString: function(nogos) {
         var s = '';
         for (var i = 0, polyline, vertices; i < nogos.length; i++) {
-            polyline = nogos[i];
-            vertices = polyline.getLatLngs();
-            for (var j = 0; j < vertices.length; j++) {
-                if (j > 0) {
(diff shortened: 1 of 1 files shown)


nrenner:
Just saw this one again - don't know what to comment on if it's not functional yet. But if it seems to solve the issue we should go for it?

Phyks:
Hi,

Sure, I was working on the profile tweaking ui lately and let this one drift a bit aside. I'll try to fix it and have a working mapbox/polyline implementation soon.

Still a WIP then.

bagage:
@Phyks FYI we are talking about too-big nogos limitation on #405 - maybe it can add extra motivation on that PR that should address that :). Or maybe you do not plan to work on it anymore?

Phyks:
Hi @bagage, thanks for the heads up, this one completely slipped out of my sight...

I'm not sure to have much free time to commit into this one soon. If anyone is interested in taking over, please feel free to do so!

bagage:
Continuing on #417 then, thanks @Phyks!

Issue #223, opened 2019-08-01
Let user export GPX with any turnInstructionStyle


Phyks:
Hi,

I've recently discovered that BRouter supports different GPX export format, https://github.com/phyks/brouter/blob/5a1f7dccedc0c97f21166f0b661d9f562b8f7829/misc/profiles2/trekking.brf#L17. 

Depending on the `turnInstructionMode` variable, this can be either no turn instructions, locus, OsmAnd etc. What about letting users overloading this when downloading the GPX trace from BRouter-web, instead of having to dig into the profile?

Actually, this might be something worth discussing at the BRouter level. The profiles mix global variables (used and defined across any profiles such as `turnInstructionMode` or `considerTurnRestrictions`), local variables (customizing the behavior of a particular profile) and auxiliary variable (for readability of the profiles).

It could be very useful to have a nice UI for at least the BRouter-global ones (and maybe the local ones, provided they can be parsed easily from the profiles, which is not currently the case). This would probably improve BRouter-web accessibility and ease of use a lot! :)

Best,

nrenner:
@abrensch mentioned that the BRouter Android intent interface would already support profile parameters, but I don't know any details and what it would take to make them passable from BRouter-Web and override the profile default values.

Also not sure about the UI. Would that be a generic one parsing available parameters from the current profile? Or fixed inputs where it makes sense, e.g. `turnInstructionMode` in the export dialog. Or a combination of both?

The primary goal for having profile parameters would be to get rid of profile variants in the dropdown (trekking ignore cycleroutes/steep/no ferries/no steps/...) and use checkbox options instead.

abrensch:
There are some undocumented URL-params that I use for an e-mobility showcase:

see https://github.com/abrensch/brouter/blob/master/brouter-server/src/main/java/btools/server/RouteServer.java

- timode : the turn-instruction-mode
- heading: the current moving direction
- profile:xxx : overwrite parameter xxx *** of the kinematical model (=car routing) ***

So that' currently only a hack and profile:xxx does not overwrite params of the "StdModel", and there are other hacks, e.g. does profile:vmax at the same time requests the speed-limit-profile.

But if that should become a feature I could of course remove the hacks and make it more consistent.

regards, Arndt

Phyks:
> Also not sure about the UI. Would that be a generic one parsing available parameters from the current profile? Or fixed inputs where it makes sense, e.g. turnInstructionMode in the export dialog. Or a combination of both?

[...]

PR #235, opened 2019-08-20
First PoC for UI for customizing profile

Phyks:
See #223 

Very first PoC, still quite some work to do on UI etc.

Screenshot_2019-08-20 Client web BRouter

What do you think about this?

**EDIT**: TODO

- [x] boolean variables should either be checkboxes or yes/no radio
   * probably parsable if default value is true or false or line including comment has
        ```
        1 # set to 0
        0 # set to 1
        ```

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -188,6 +188,10 @@ input#trackname:focus:invalid {
     cursor: auto;
 }
 
+#profile_editor {
+    display: none;
+}
+
 #profile_buttons {
     padding-top: 4px;
 }
--- a/css/style.css
+++ b/css/style.css
@@ -426,12 +426,35 @@ table.dataTable.display tbody tr.even:hover {
  * leaflet-sidebar-v2
  */
 
-.leaflet-sidebar-pane#tab_profile,
 .leaflet-sidebar-pane#tab_data,
 .leaflet-sidebar-pane#tab_itinerary {
     /* Full height for content with inner scrolling,
        overrides scroll fix for long content in Firefox */
     height: 100%;
+}
+
+.leaflet-sidebar-pane#tab_profile {
+    min-height: 100%;
+}
+
+.leaflet-sidebar-pane#tab_profile .form-group {
+    margin-bottom: 5px;
+}
+
+.leaflet-sidebar-pane#tab_profile label {
+    font-weight: 700;
+    margin-bottom: 0;
+}
+
+.leaflet-sidebar-pane#tab_profile input[type='checkbox'] {
+    margin-right: 5px;
+    vertical-align: middle;
+}
+
+.leaflet-sidebar-pane#tab_profile .help-block {
+    display: block;
+    color: #737373;
+    margin-bottom: 0.5rem;
 }
 
 .leaflet-sidebar-content {
--- a/index.html
+++ b/index.html
@@ -554,8 +554,8 @@
                         <a
                             href="#tab_profile"
                             role="tab"
-                            data-i18n="[title]sidebar.custom-profile.title"
-                            title="Custom profile"
+                            data-i18n="[title]sidebar.customize-profile.title"
+                            title="Customize profile"
                             ><i class="fa fa-wrench"></i
                         ></a>
                     </li>
@@ -643,29 +643,21 @@
                             <h1 class="leaflet-sidebar-header">
                                 <span class="leaflet-sidebar-close"><i class="fa fa-caret-right"></i></span
                                 ><span class="leaflet-sidebar-expand"><i class="fa fa-expand"></i></span
-                                ><span data-i18n="sidebar.custom-profile.title">Custom profile</span>
+                                ><span data-i18n="sidebar.customize-profile.title">Customize profile</span>
                             </h1>
                             <form class="flexcolumn flexgrow">
                                 <div id="profile_params_container">
                                     <div id="profile_params"></div>
                                     <div class="form-group" id="profile_buttons">
-                                        <button
-                                            id="upload"
-                                            type="button"
-                                            class="btn btn-primary btn-sm"
-                                            data-uploading-text="Uploading…"
-                                        >
-                                            <span class="fa fa-cloud-upload"></span>
-                                            <span data-i18n="sidebar.profile.upload">Upload</span>
-                                        </button>
-                                        <button
-                                            type="button"
-                                            class="btn btn-info btn-sm pull-right"
-                                            id="profile_advanced"
-                                        >
+                                        <button type="button" class="btn btn-info btn-sm" id="profile_advanced">
                                             <span data-i18n="sidebar.profile.switch_advanced"
                                                 >Switch to advanced editor</span
                                             >
+                                        </button>
+
+                                        <button id="save" type="button" class="btn btn-primary btn-sm pull-right">
+                                            <span class="fa fa-cloud-upload"></span>
+                                            <span data-i18n="sidebar.profile.save">Save</span>
                                         </button>
                                     </div>
                                 </div>
@@ -700,6 +692,10 @@
                                             ><span class="fa fa-question"></span>
                                             <span data-i18n="sidebar.profile.help">Help</span></a
                                         >
+
+                                        <button type="button" class="btn btn-info btn-sm" id="profile_basic">
+                                            <span data-i18n="sidebar.profile.switch_basic">Switch to basic editor</span>
+                                        </button>
                                     </div>
                                 </div>
                             </form>
--- a/index.html
+++ b/index.html
@@ -646,37 +646,61 @@
                                 ><span data-i18n="sidebar.custom-profile.title">Custom profile</span>
                             </h1>
                             <form class="flexcolumn flexgrow">
-                                <div id="profile_params"></div>
-                                <textarea
-                                    class="form-control flexgrow"
-                                    id="profile_upload"
-                                    spellcheck="false"
-                                    wrap="off"
-                                    data-i18n="[placeholder]sidebar.profile.placeholder"
-                                    placeholder="Write your custom profile here."
-                                ></textarea>
-                                <div id="profile_message"></div>
-                                <div class="form-group" id="profile_buttons">
-                                    <button
-                                        id="upload"
-                                        type="button"
-                                        class="btn btn-primary btn-sm"
-                                        data-uploading-text="Uploading…"
-                                    >
-                                        <span class="fa fa-cloud-upload"></span>
-                                        <span data-i18n="sidebar.profile.upload">Upload</span>
-                                    </button>
-                                    <button id="clear" type="button" class="btn btn-secondary btn-sm">
-                                        <span class="fa fa-eraser"></span>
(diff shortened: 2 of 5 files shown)


nrenner:
Some thoughts:

- boolean variables should either be checkboxes or yes/no radio
  - probably parsable if default value is `true` or `false` or line including comment has
    - `1   # set to 0`
    - `0   # set to 1`
- internal flags like `validForBikes` probably can be excluded
- this should get its own tab, separate from the editor
  - I added the sidebar to get full height for the editor
  - more user friendly, no need to confront user with profile code and editor UI when setting parameters
- turnInstructionMode would really make sense in the export dialog (in addition?)
  - generally I wouldn't mind having hard-coded inputs for parameters that are also used in server code (outside of profile code)
  - but context-aware: only for supported formats and when currently selected profile has that variable (?)

Phyks:
> internal flags like validForBikes probably can be excluded

Sure, but this would require either maintaining a separate list of such internal variables or having a different notation for these in the BRouter profile.

> this should get its own tab, separate from the editor 

My idea was to have this as an "easy interface" for editing a profile and then a "toggle expert mode" link which would show the current textarea instead.


My goal with this would be, at least in a first time, to get rid of the many variants of some profiles. For instance, the many "trekking" alternatives which actually only differs by a few variables.

nrenner:
> Sure, but this would require either maintaining a separate list of such internal variables 

Yes, why not. Or ignore variables without comment. I guess it doesn't make sense to turn off `validFor*` modes, so why present it to the user? 

> My idea was to have this as an "easy interface" for editing a profile and then a "toggle expert mode" link which would show the current textarea instead.

Just to make sure we are talking about the same thing: 

This sounds like you plan to automatically edit the profile text and replace the default values with the inputs in the client and send it (via "upload" button?) as a customized profile to the server (without any changes in the server)?

What I would like to have is sending the inputs as URL parameters to the server where they override the default values after parsing (don't know where exactly this is/should be done in the server).

Phyks:
> Yes, why not. Or ignore variables without comment. I guess it doesn't make sense to turn off validFor* modes, so why present it to the user?

Indeed, but this would also mean we lose variables such as [`uphillcost`](https://github.com/abrensch/brouter/blob/master/misc/profiles2/trekking.brf#L25).


> This sounds like you plan to automatically edit the profile text and replace the default values with the inputs in the client and send it (via "upload" button?) as a customized profile to the server (without any changes in the server)?

This is the quick and easy going way.

> What I would like to have is sending the inputs as URL parameters to the server where they override the default values after parsing (don't know where exactly this is/should be done in the server).

I agree this would be definitely better. I can look into updating the backend (server) code as well.

Phyks:
Another question on this is what we should do with the BRouter global variables defined in https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/RoutingContext.java#L127-L188. Most of the time, the default value is not even listed in the profiles.

1. Should the profiles include them with the default values and a `%` comment when they can be of use (`totalMass` etc for instance for bikes profiles)?
2. Should we somehow import them in BRouter-web, no matter the profile?

Phyks:
@nrenner I gave a try at an extended Locus syntax. This PR is updated accordingly. The updated `trekking.brf` profile is available at https://gist.github.com/Phyks/fb9136bdfc81124540ebfcb067048bec (all relevant variables are commented for Locus / BRouter-web).

Current UI is:

Screenshot_2019-09-02 BRouter web client

It still requires a bit of work but if we agree on the backend (profiles) syntax, I'll finish the editing and write a PR to BRouter in parallel of working on the UI.

bagage:
This is looking good, keep going @Phyks :)! I'd just want to put emphasis on [@zossebart comment](https://github.com/nrenner/brouter-web/issues/223#issuecomment-526832098) about translations: maybe the simplest solution is to use the variable name as translation key, eg `ignore_cycleroutes.description` would match `Set to true for better elevation results` and so on?

nrenner:
Thanks for the update, nice progress!

Wanted to say the same about translations, we could also leave off the display name (you already did?) and set it in translations only, like we do for profile names.

Phyks:
> maybe the simplest solution is to use the variable name as translation key, eg ignore_cycleroutes.description would match Set to true for better elevation results and so on?

What about having name and description as currently done in https://github.com/Phyks/brouter/blob/fc13c5c1594575cc92b912536c7fa9b77bd72d1c/misc/profiles2/trekking.brf#L12 with English (just as currently done in BRouter). Apps would be free to use either the provided name and description or use the name key for translations.

In BRouter-web, this would mean, taking https://github.com/Phyks/brouter/blob/fc13c5c1594575cc92b912536c7fa9b77bd72d1c/misc/profiles2/trekking.brf#L12 as an example,

* If `profileParameters.allow_steps.name` key exists in i18n files, it will be used for the name. Otherwise, `allow_steps` will be shown.
* If `profileParameters.allow_steps.description` key exists in i18n files, it will be used for the description. Otherwise, `Set to false to disallow steps` will be shown.

I'll edit the PR accordingly.

nrenner:
> Another question on this is what we should do with the BRouter global variables defined in https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/RoutingContext.java#L127-L188. Most of the time, the default value is not even listed in the profiles.

I personally would tend to keep this as user friendly as possible and reduced to a minimum of parameters that are likely to be changed in everyday use / from route to route and are ideally self-explaining.

So I would only use those currently in the profiles, or just a selection of them (e.g. probably without `uphillcost`, etc.).

Unless we either have an expert mode or have groups of parameters (like in your profile or the RoutingContext code) and show only the most important and other groups collapsed (maybe as a second step later).

Phyks:
> So I would only use those currently in the profiles, or just a selection of them (e.g. probably without uphillcost, etc.).
> Unless we either have an expert mode or have groups of parameters (like in your profile or the RoutingContext code) and show only the most important and other groups collapsed (maybe as a second step later).

I added the relevant parameters in my update of the trekking profile (https://github.com/Phyks/brouter/blob/fc13c5c1594575cc92b912536c7fa9b77bd72d1c/misc/profiles2/trekking.brf#L8-L40). This is also to try to improve on https://github.com/abrensch/brouter/issues/184 a bit.

We could have groups in the future, parsed from the sections delimited by comments in the profile for instance. Anyways, if we start having translations of parameters names and descriptions in BRouter-web, we could also have this grouping stored within BRouter-web.

In short, my opinion is the edit profile tab is currently for quite advanced users. So, showing all parameters in the first version would probably not be so bad, since we are moving from an advanced users screen to a semi-advanced users screen :) We might refine this afterwards, for sure.

Phyks:
I updated the PR, the user-friendly edit interface should now be styled and working, as well as the switch to the advanced editor.

Here are a few screenshots:

ui

ui2

If you want to give it a try, you should use this PR on BRouter: https://github.com/abrensch/brouter/pull/189.

Happy to take any feedback on this!

bagage:
I cannot make it work yet, with your PR trekking.brf:

> java.lang.IllegalArgumentException: ParseException at line 68: unknown lookup name: bicycle_road

nrenner:
Looking good!

There are a few issues with profile state management and syncing, but I don't know if this is supposed to be fully working already?

Phyks:
Thanks! 

> I don't know if this is supposed to be fully working already?

It should :)

> There are a few issues with profile state management and syncing

Happy to have more details on this. I tested it but probably missed some cases. Thanks!

nrenner:
The (advanced) editor is not synchronized:
1. uncheck "allow_steps", "Save"
2. "Switch to advanced editor"
   -> `assign allow_steps = true` (should be false)
3. "Switch to basic editor"
  -> "allow_steps" is checked again (should be unchecked)

The JavaScript error I see in the console when switching profile and back might be caused by this.

Phyks:
Should be rebased (for conflicts) and fixed.

Phyks:
BRouter counterpart has been merged! :)

nrenner:
Seems like the errors in the console weren't related, I still get them:

`Error getting profile from "profiles2/custom_....brf"` (Profile.js:51)

1. change a parameter in basic editor and "Save"
2. switch profile to "Trekking bike" and back to "Custom"

The request to get the profile text shouldn't be executed for a custom profile, I think this condition shouldn't be true:
https://github.com/nrenner/brouter-web/blob/179f1eedf0430a6a2a5b8499180c43f8d2896347/js/control/Profile.js#L35

There probably needs to be some additional state mangagement for the basic editor.


`Uncaught TypeError: Cannot read property 'replace' of undefined` (Profile.js:110)

3. change a property again and "Save" (after 1. and 2. above)

The same error occurs when app called with a route:

1. http://localhost:3000/#map=18/48.86226/2.33030/osm-mapnik-german_style&lonlats=2.330746,48.86241;2.330974,48.862333
2. change parameter and "Save"

Phyks:
> The same error occurs when app called with a route:

Well caught, this should be fixed now!


> The request to get the profile text shouldn't be executed for a custom profile, I think this condition shouldn't be true:

I agree, but I really don't know what would be the user-friendly behavior here. Current behavior in BRouter-web is not optimal either :/

I pushed a second commit which should fix the issue, I think. Custom profiles are now cached upon upload so that it is easy to recover them!

nrenner:
The fixes look good to me, thanks! Nice idea with the caching.

One thing left is that the profile view/editor is too hidden now in my opinion. You can't see it's there until you scroll all the way down to the end of the options and read the button.

Don't know if users actually use it that way, but I see it not only as an editor for advanced users, but also for viewing the current profile:
- provides a description of the profile in the header comment
- offers some transparency about what the router actually does, 
  e.g. allows to check if and how a specific tag is considered
- shows off the unique feature of a fully customizable profile at runtime

My suggestion would be using [Bootstrap tabs](https://getbootstrap.com/docs/4.3/components/navs/#javascript-behavior) instead of buttons to switch:
- would be the top where you expect navigation and immediately visible
- leaves space for more buttons (plan to add "Save", "Load" and "Share")
- tab UI element implies switching, no need to explain with "Switch to ..."
- maybe name it "Options" and "Profile" instead of "basic editor" and "advanced editor"?
- and rename "Save" to "Apply", because also applicable to "proper" profile parameters later?

What do you think? If you prefer, I could also do this after merging.

Phyks:
Hi,

Thanks for the feedback on this! This sound better in terms of UI indeed.

I'll try to have a look next week, but I'll not have time before Tuesday. So, feel free to merge and edit this if you have time before :)

nrenner:
I'm going to have a look at adding the Bootstrap tabs, unless you already started?

Phyks:
> I'm going to have a look at adding the Bootstrap tabs, unless you already started?

Did not start anything already.

Thanks for the merge and truly sorry for my late reply, I've had far fewer free time than expected lately :/

nrenner:
Implemented as suggested, see #260

Phyks:
Thanks!

Stapawe:
After looking on screenshots, will be in this UI the most important thing - highway costs? Thank you for your great job, especially in the last year or so.

Issue #234, opened 2019-08-19
yarn watch is failing after first run


Phyks:
If running `yarn watch`, the code is built once. Then, on first change on a file, the rebuild is done correctly. After this, it seems subsequent builds are not working anymore. Not sure what is happening.

Phyks:
Seems that on a different machine (mac instead of linux), it actually works fine :/

nrenner:
Maybe gulp has some verbose/debugging output you could turn on?

nrenner:
I needed to increase the inotify file watch limit on Ubuntu 18.04 - but I got this explicit error message from Gulp watch (and similar from VS Code):
`Error: ENOSPC: System limit for number of file watchers reached, watch '...'`

https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers#the-technical-details

bagage:
I'm trying to get more logs but `gulp -LLLL watch` does not log anything.

But for me, the issue only happens when running gulp via Intellij IDEA - which consumes a lot of memory and also a lot of file watchers, so maybe I'm getting some issue around that (even if I raised the number to `524288`). Still investigating…

Phyks:
I could no longer reproduce it today (before your PR). Not sure what was happening. Feel free to close this one if you cannot reproduce either.

bagage:
I'll keep going on #222 and if I cannot reproduce it i'll close this one.

PR #238, opened 2019-09-02
Replace watch by serve command with livereload

bagage:
Still having #234 issue which makes this feature a bit useless for me right now, but I'll look at it (then I'll go back to #222 :)).

Reviews are always welcome :).


bagage:
OK I was missing a `done()` call. This is now working correctly but I doubt I fixed #234 yet.

PR #239, opened 2019-09-20
Display a popup on init for France users

bagage:
In France there is currently a poll for bike users, created by the biggest French bikers association. This poll is about bike usability in big cities and is going to last until end of November. 

Please let me know if you think Brouter is not the place to communicate about it :).

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -15,6 +15,28 @@
         }
     }
 
+    function displayWelcomePopup(map, search) {
+        if (BR.Util.localStorageAvailable()) {
+            var alreadyShown = localStorage.getItem('welcomePopupFub') == 1;
+            var isActive = new Date() < new Date(2019, 11, 1);
+            if (!alreadyShown && isActive) {
+                var nominatim = search.options.geocoder.options.next;
+                nominatim.reverse(map.getCenter(), map.getZoom(), function(result) {
+                    if (result && result[0] && result[0].name.indexOf('France') !== -1) {
+                        bootbox.confirm({
+                            title: 'Parlons vélo - enquête usagers',
+                            message:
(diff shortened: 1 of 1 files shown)


nrenner:
I'm a bit torn between supporting a good cause to improve bicycle infrastructure and my general aversion to the ever increasing demand for attention that keeps distracting you from doing what you actually want, and I would lean towards not having any campaign ads on our site.

bagage:
I get the mixed feelings @nrenner. Actually I like what openstreetmap.org does: some banners are below search bar on the left:

image

Maybe they removed "local events" but at a time they would appear there in place of the welcome card.

In the mean time, the popup will appear only once per French user (if they click OK button - I can change for popup dismissal).

bagage:
closing this for now, maybe i'll think for something else next year :).

nrenner:
Sorry, wanted to respond again earlier, still have a backlog of things to do and to comment on.

I have strong opinions about agressive ads and use an ad blocker to defend against them. It would be contradicting to support adding a popup - which is agressive - to our site. Yes, it's targeted to the relevant group and shown only once, but still. osm.org conference banners are better, but still obtrusive and therefore [blocked](https://github.com/easylist/easylist/pull/900) by ad blockers.

But the form is not really my point. The purpose of any kind of ad is to distract and get attention. The more I think of it, the more I believe there should be spaces where the user is left alone both from commercial ads and calls to action for surveys, petitions and the like, and that sites I contribute to should be one of those.

I get that it would be a nice way to target a relevant user group, but I think communication channels (like Twitter, Facebook, ...) are better places where you expect to get informed about such things.

bagage:
I do understand your point of view and I'm totally OK with it. There are indeed enough others channels to do these communications, I thought in the first place that brouter could be an ideal way of reaching the audience, but actually it is not :). I'll revert the changes from my fork too btw.

PR #240, opened 2019-10-01
Feature/load gpx track

matzepan:
Added the ability to load tracks (kml, gpx, geojson, json) using leaflet-filelayer and display them as new Overlay layer.
Addresses: https://github.com/nrenner/brouter-web/issues/30

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,3 +180,7 @@ Copyright (c) 2018 Norbert Renner and [contributors](https://github.com/nrenner/
     Copyright (c) 2014-2016 GitHub, Inc.; [MIT License](https://github.com/github/fetch/blob/master/LICENSE)
 -   [Promise Polyfill](https://github.com/taylorhakes/promise-polyfill)  
     Copyright (c) 2014 Taylor Hakes, Copyright (c) 2014 Forbes Lindesay; [MIT License](https://github.com/taylorhakes/promise-polyfill/blob/master/LICENSE)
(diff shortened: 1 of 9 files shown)


bagage:
Hey @matzepan, welcome aboard! Thanks for your contribution, it is one of the most asked-for feature. From what I've seen it's looking good, albeit a *minor* issue: it looks like tracks are declared as interactive layers but it seems they are not:

Capture d’écran vidéo de 01-10-2019 19:27:32_cropped

Good work!

matzepan:
@bagage this is also a feature that I wanted badly.
I changed the loaded track to be not interactive.
Thank you for your feedback.

Phyks:
What about keeping the loaded track in a URL parameter for easy sharing of the full view? For instance by encoding the polyline with `mapbox/polyline` in a `track` (`tracks`?) GET parameter similarly to https://github.com/nrenner/brouter-web/pull/233 (which I should finish by the way)?

matzepan:
@Phyks I don't know if it makes sense to put these into the URL. Especially if this means blowing up the URL length.

Phyks:
It makes sense to have it to easy share the view with someone else / another computer. But indeed it probably does not make sense to have it in the URL and risk blowing the URL length.

Maybe BRouter-web should have something like Overpass to shorten links and share easily (including custom profiles, etc).

nrenner:
Great contribution, thanks!

Issue #19, opened 2015-06-17
Feature: color route segments according their "quality"


DrVanScott:
It would be a nice feature for route planning if the ways somehow reflect their "quality". Quality could be some of

 * slope
 * way condition (a index combining 'surface', highway, isBike, ...)
 * cost

poutnikl:
quite straighforward could be to follow OSMAnd way to graphically display tags
tracktype=*
surface=*
smoothness=*
eventually olso
mtb:scale=*

It may be matter rather of map providers,  but overlay layers can help much.

nrenner:
Certainly a feature I personally would also like to have.

poutnikl:
This can be distinguished and implemented  as 2 independent things:

Marking based on OSM tagging and related info, (inspired by OSMAnd ) like 
system of solid/dash/dot marking for track grades, 
track background colour for surface=* , 
marking road conditions or MTB/SAC scale, 
several sharp arrows for big incline. etc

Marking based on profile cost, represented by colour of route. 
Something similar to colouring of routes, according to above sea level, present on some GPX processing websites.

drazde:
I found very useful the OsmAnd layout for MTB. When I'm on mountain on a path i could take decision very quickly.
To plan a route I would like the same clear information (path colored green,blue,red,grey)


![Screenshot_20231024_083832_OsmAnd2](https://github.com/nrenner/brouter-web/assets/2643448/16be3254-fca0-4227-85c5-c211aafcba54)

Issue #229, opened 2019-08-10
i18n: let users overload the initial setting


Phyks:
Hi,

What about letting users overload the default i18n value (based on browser locale I think)? I have two typical examples in mind:

1. When developping BRouter, it would be an easy way to ensure having an English locale easily (and then using the "default" translations).
2. A user reported to me that he was thinking brouter.de/brouter-web was only available in German as he was seeing the German version and no way to change the language. Therefore, he was reluctant to passing the link to some French friends, thinking they would not understand the interface.

Best,

bagage:
You can use `?lng=en` in the URL like https://brouter.damsy.net/latest/?lng=de#map=12/45.0574/5.1955/standard to force the langage. In this case it is persisted in local storage.
But we could add a language picker indeed,sounds easy to implement

Phyks:
Oh, I did not know about the `lng` GET parameter, thanks for the tip!

:+1: for having a language selector.

nrenner:
We were undecided about the language selector, but as your example shows, I guess it would at least make it clear that the app is translatable and what languages are available.

Looks like switching can be done with [`i18next.changeLanguage`](https://www.i18next.com/overview/api#changelanguage).

Would be nice to have a way to get the list of available languages without requiring a release, e.g. from `config.js` or a separate file generated by the build or `transifex-pull` script.

tvaknin:
Hi, Isn't this a fundamental requirement? I think it would help a lot.

PR #242, opened 2019-10-01
Feature/color route segments

matzepan:
Added a overlay which reflects the quality of the route based on either:
* cost
* altitude
* incline
Partially adresses https://github.com/nrenner/brouter-web/issues/19 and could be extended by using https://github.com/hgoebl/Leaflet.MultiOptionsPolyline to overlay the road types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,3 +184,5 @@ Copyright (c) 2018 Norbert Renner and [contributors](https://github.com/nrenner/
     Copyright (c) 2012 Makina Corpus; [MIT License](https://github.com/makinacorpus/Leaflet.FileLayer/blob/master/LICENSE)
 -   [togeojson](https://github.com/mapbox/togeojson)  
     Copyright (c) 2016 Mapbox All rights reserved.; [2-clause BSD License](https://github.com/mapbox/togeojson/blob/master/LICENSE)
+-   [Leaflet.hotline](https://github.com/iosphere/Leaflet.hotline)  
+    Copyright (c) 2015, iosphere GmbH, Jonas Coch; [Leaflet.hotline](https://github.com/iosphere/Leaflet.hotline/blob/master/LICENSE)
--- a/js/index.js
+++ b/js/index.js
@@ -199,6 +199,8 @@
             requestUpdate: requestUpdate
         });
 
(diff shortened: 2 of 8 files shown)


bagage:
I didn't looked at the code yet and wanted to give it a try, looks also good! I'm not sure if `cost` should be the default one (over incline), but 3 visualizations are complementary.

The coloration doesn't go away when you refresh route (eg. if you switch between profile, you end up with the old colored profile and the new pink profile, until the route is fully calculated). Is that a bug or a feature?

Capture d’écran vidéo de 01-10-2019 19:48:24_cropped

matzepan:
@bagage this is just the way components get notified when a total reloading of the track is done. Like the elevation chart.
But you're right the cost function currently might be most useful for me because I want to see where the algorithm shows the most costly paths. I changed it to be incline first.

Next thing would be to maybe switch to https://github.com/hgoebl/Leaflet.MultiOptionsPolyline and get this to a separate tab to the right similar to:
image

Phyks:
> Next thing would be to maybe switch to https://github.com/hgoebl/Leaflet.MultiOptionsPolyline and get this to a separate tab to the right similar to

There was the idea to have a separate tab for BRouter-web configuration (for instance for #229). Maybe this would be a good place to put the options for the path overlay color as well?

matzepan:
@bagage can we switch to gulp-uglify-es to support ecmascript 6?

bagage:
Totally that would be cool! Is there any reason not to @nrenner?

nrenner:
I opened a new issue for the ES 6 question: #249.

Great to see many pull requests! Will try to look through them in the next days.

nrenner:
Nice! I do have some comments, but will open a follow-up issue after merging.

Just a minor wish right now: can we move the button just above the route [opacity slider](https://github.com/nrenner/brouter-web/blob/90dfa5b7f33054e61ae258e98188eac46954b4a3/js/index.js#L276) (by moving the `.addTo(map)`)? It's less important than the `drawToolbar` and fits thematically, as both change the route appearance.

nrenner:
Thanks!

nrenner:
A follow-up issue: #262

Issue #169, opened 2019-04-15
distance marker (adjustable)


radlerandi:
I use brouter-web since some years to plan my bike trips, and i like it very much! thanks to your efforts!

It would be very helpful if there would be a distance marker, for example every 50 or 100 km (probably adjustable).

Also very helpful would be a button to split a track in 2 parts and get 2 perma links, that i can open in a new browser window.

bagage:
Markers are currently used as a way to adjust the route generated by brouter engine. Maybe distance markers could be added on top of the route, but without interactivity enabled ("read only")?

PR #243, opened 2019-10-02
Feature/Added distance markers to the routing path

matzepan:
Adding distance markers each 5km.
Currently this is not configurable.

Addresses: https://github.com/nrenner/brouter-web/issues/169


bagage:
Regarding the fact that this is not (yet) configurable, I'll create another issue for BRouter-web general configuration. I think having a settings for it there makes sense?

nrenner:
Thanks! Nice improvements to the original plugin, might be worth making a PR for your changes?

PR #245, opened 2019-10-02
gulp serve must depends on debug first

### LINKED ISSUE #244: gulp serve opens browser before style has been computed
when running `gulp serve` the browser is opened before the CSS has been built, and so you end on a broken page.
You have to manually refresh the page to make it OK.

bagage:
Fixes #244


nrenner:
Thanks, haven't noticed that yet.

PR #251, opened 2019-10-05
Add CyclOSM as usable layer

Phyks:
Hi,

This adds the [CyclOSM](https://www.cyclosm.org/) layer as a possible layer offered.

So far, it is hidden in the tree, to provide an easy way to use it for users who want, but not put too much pressure on the servers. We can see afterwards, depending on usage and servers capacity to eventually move it to the default list or to be the default background.

@nrenner One thing I'd like to do (but I could not find how to do it) is use it as the default background for French users (instead of the osm-fr style). Since both osm-fr and cyclosm are currently hosted on the same infrastructure there should not be any impact in terms of performance to do this switch and the CyclOSM render is (I hope) way better for cycling than the osm-fr one.

Best,

### DIFF
--- a/layers/config/config.js
+++ b/layers/config/config.js
@@ -10,7 +10,7 @@ BR.confLayers.defaultBaseLayers = [
 // worldwide monolingual layers to add as default when browser language matches
 BR.confLayers.languageDefaultLayers = [
     'osm-mapnik-german_style',
-    'osmfr',
+    'cyclosm',
     '1021' // sputnik.ru
 ];
 
--- a/layers/config/overrides.js
+++ b/layers/config/overrides.js
@@ -227,6 +227,11 @@ BR.confLayers.getPropertyOverrides = function() {
         'mapaszlakow-routes': {
             'nameShort': 'Routes PL',
             'mapUrl': 'http://mapaszlakow.eu/#{zoom}/{lat}/{lon}'
-        }
+        },
+        'cyclosm': {
+            'language_code': 'fr',
+            'nameShort': 'CyclOSM',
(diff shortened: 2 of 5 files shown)


nrenner:
osm-fr for French users is configured in:

https://github.com/nrenner/brouter-web/blob/179f1eedf0430a6a2a5b8499180c43f8d2896347/layers/config/config.js#L13

and

https://github.com/nrenner/brouter-web/blob/179f1eedf0430a6a2a5b8499180c43f8d2896347/layers/config/overrides.js#L78

So as a bit of a hack you could replace `osmfr` in `config.js` and add a `cyclosm` entry in `overrides.js` with a fake `'language_code': 'fr',` property.

Phyks:
> So as a bit of a hack you could replace osmfr in config.js and add a cyclosm entry in overrides.js with a fake 'language_code': 'fr', property.

Thanks, I pushed an extra commti which does this. Not sure why, but this does not work locally (nor the default to `osm-fr` with the current codebase and French locale). :/

nrenner:
> Not sure why, but this does not work locally (nor the default to osm-fr with the current codebase and French locale).

- works for me when calling with http://localhost:3000/?lng=fr
- layers and the config files need to be built first
- check language detection in console:
  - detected language hierarchy: `i18next.languages`
  - stored detection: `localStorage.i18nextLng`
  - delete stored detection: `localStorage.removeItem('i18nextLng')`
- the matching code is in [LayersConfig._addLanguageDefaultLayer](https://github.com/nrenner/brouter-web/blob/d31be926d2dfc8e5402f5fd5c40e81a167275009/js/LayersConfig.js#L80-L106)

PR #253, opened 2019-10-17
Use custom markers for start and end

### LINKED ISSUE #124: Brouter doesn't understand one-way cycleways.
Brouter doesn't seem to understand ways like [this one](https://www.openstreetmap.org/way/584971528), which have a cycleway only in one direction (in this case, only in the Southbound direction). It is suggesting using this road when traveling North to South, even though it doesn't actually have any cycling facilities at all (and there are other nearby Southbound options that do).

[here](http://h2096617.stratoserver.net:443/brouter?lonlats=-74.039408,40.715452|-74.015858,40.713972&nogos=&profile=custom_1530022627267&alternativeidx=0&format=geojson) is an example route that shows this problem in practice.

bagage:
Closes #124 

Thanks @nrenner! Icons could probably be improved though, I used the ones from https://github.com/mpetazzoni/leaflet-gpx but I think a designer could do something even better :).

image


bagage:
I actually end up using the same icons as for POIs:

image

bagage:
Yes indeed I wasnt sure myself about the bookmark icon, glad you pointed it out too 😅. Will update it

nrenner:
Thanks! This has been one of the still missing features to become a "proper" routing client and is one more step towards a v1.0.

Issue #211, opened 2019-06-20
Overlay selector is counter-intuitive


Phyks:
Hi,

It seems the overlay selector is broken. Indeed, when going to the layers panel, expanding the list and clicking "Mapillary coverage", the line gets selected but the checkbox is never ticked. The overlay is actually rendered without overlaying apparently.

![2019-06-21-004326](https://user-images.githubusercontent.com/3856586/59886034-b7026400-93bd-11e9-9d1e-07c45b14305e.png)

When clicking the checkbox, nothing happens.

![2019-06-21-004333](https://user-images.githubusercontent.com/3856586/59886043-bb2e8180-93bd-11e9-9c8f-e230c85ee9ec.png)

nrenner:
This is how it's supposed to work:
- click on the name shows a preview of that layer without any other layers (active layers turned off)
- the checkbox adds (or removes) the layer to the layer switcher

So despite looking similar with the checkbox, it's not an extended layer switcher - it configures the layer switcher. I guess this somehow needs to be made more clear.

Phyks:
> click on the name shows a preview of that layer without any other layers (active layers turned off)

Ok! This works fine for me (although this is not very intuitive). I'd tend to prefer not changing the usual behavior of clicking a label to tick the checkbox (especially on touch screens). Another option (maybe clearer from UX point of view) would be to have a magnifying glass icon next to the item to toggle the preview mode?

![Capture d’écran 2019-06-21 à 11 39 01](https://user-images.githubusercontent.com/3856586/59913795-2e191600-9419-11e9-9081-cfc62ed809ad.png)

> the checkbox adds (or removes) the layer to the layer switcher

When ticking the checkbox, the overlay layer is not shown on the map for me (checked with different overlays, this is not a mapillary coverage bug). I can reproduce on http://brouter.de/brouter-web/.

No error in the console, no network calls made (so the layer is somehow not fetched) :/

This is only the case for the "More" overlays, such as Mapillary coverage or Waymarked Trails MTB. The default overlays (Waymarked Trails Hiking/Cycling etc) do work fine.

nrenner:
Ticking the checkbox does not show the layer, it only adds it to the layer switcher list. To activate you need to tick it in the layer switcher (tha part above the buttons).

Phyks:
> Ticking the checkbox does not show the layer, it only adds it to the layer switcher list. To activate you need to tick it in the layer switcher (tha part above the buttons).

[...]

PR #263, opened 2019-11-28
Add hints and feedback for optional layers tree usage

nrenner:
An attempt to make usage of the optional layers tree more clear, see #211, while keeping the behavior for now:

- help message on top (using button tooltip): "Add or remove optional layers"
- icon tooltips as hints and feedback for add/remove and preview actions
- "Preview" label on the map
- lower opacity and default cursor for disabled root node checkboxes

@Phyks what do you think?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -186,9 +186,19 @@ input#trackname:focus:invalid {
 #message {
     position: absolute;
     margin: 10px 46px; /* 10 + 26 + 10 */
-    z-index: 3000;
+    z-index: 3001;
     font-size: 1rem;
     cursor: auto;
+}
+#preview {
+    position: absolute;
+    top: 10px;
+    padding: 0.5rem 1rem;
+    border-radius: 4px;
+    z-index: 3000;
+    font-size: 1.5rem;
+    background-color: rgba(0, 116, 217, 0.6);
+    color: white;
 }
 
 /* Override Bootstrap tabs that set `display` from `none` to `block` when activating */
@@ -463,8 +473,10 @@ table.dataTable.display tbody tr.even:hover {
 
 .leaflet-sidebar-pane#tab_profile .help-block {
     display: block;
-    color: #737373;
     margin-bottom: 0.5rem;
+}
+.help-block {
+    color: #737373;
 }
 
 .leaflet-sidebar-content {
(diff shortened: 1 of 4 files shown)


Phyks:
Looks really nice! Can you only show the "+ / -" icon tooltips when hovering the checkbox?

nrenner:
I could, this is where I started, but apparently not the preview tooltip when only hovering the text part, as the text is not wrapped in an element:

```html
<a class="jstree-anchor jstree-checked" href="#" tabindex="-1" id="standard_anchor">
  <i class="jstree-icon jstree-checkbox" role="presentation"></i>
  <i class="jstree-icon jstree-themeicon" role="presentation"></i>
  OpenStreetMap
</a>
```

CSS select for the checkbox in the browser console:
```
document.querySelector('.jstree-default .jstree-checked .jstree-checkbox')
```

But text only seems not to be selectable in CSS (?):
[Is there a CSS selector for text nodes? - Stack Overflow](https://stackoverflow.com/questions/5688712/is-there-a-css-selector-for-text-nodes)

And I don't want to get into overriding the text rendering in JSTree right now:
[Wrap text of \<a> (jstree-anchor) into \<span> · Issue #1930 · vakata/jstree](https://github.com/vakata/jstree/issues/1930)

So I thought it would make sense and be more consistent to have them both behave the same way? And I guess always showing both is also a better immediate hint at the two different actions for an entry?

Phyks:
> But text only seems not to be selectable in CSS (?):

You probably cannot do it from within the `querySelector` call (no CSS selector for text nodes), but you might be able to post-filter it. See https://stackoverflow.com/a/10730777 for instance (if using JS here is fine).

> And I don't want to get into overriding the text rendering in JSTree right now:

:+1:

> So I thought it would make sense and be more consistent to have them both behave the same way? And I guess always showing both is also a better immediate hint at the two different actions for an entry?

In case there is an easy way to do it, I think this would be a nice improvement. But definitely :+1: for me for not spending too much time hacking around JSTree so current behavior is fine as well.

nrenner:
Hovering is now separate for checkbox and text.

As the text property accepts HTML (and I actually already use this for the language/country code prefix), wrapping it in an element was actually quite simple, no need for a plugin.

PR #272, opened 2020-01-30
Dockerfile for building and running brouter-web

saesh:
Hey @nrenner. This PR adds a multistage Dockerfile which does this:

- build brouter-web in a Docker container running node
- copies the build result into a Docker container running just Nginx making it available at port 80 by default

I have been using this Dockerfile for long time to automatically build my fork of brouter-web on Docker Hub. This makes deployment easy and fast.

Are you interested in adding it to the repo?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
 FROM node:8-alpine as build
-RUN apk add --no-cache git
 RUN mkdir /tmp/brouter-web
 WORKDIR /tmp/brouter-web
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine as build
+RUN apk add --no-cache curl unzip
+RUN curl -L `curl -s https://api.github.com/repos/nrenner/brouter-web/releases/latest | grep browser_download_url | cut -d '"' -f 4` -o brouter-web.zip
+RUN mkdir /tmp/brouter-web
+RUN unzip -d /tmp/brouter-web brouter-web.zip
+
+FROM nginx:alpine
+COPY --from=build /tmp/brouter-web/index.html /usr/share/nginx/html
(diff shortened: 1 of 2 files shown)


bagage:
@saesh I am personnally totally for it! In case @nrenner agrees too, could you add something in the readme about it?

Thanks for the contribution!

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:8-alpine as build
+RUN apk add --no-cache git
```
nrenner: Is installing Git really needed?

saesh: Good catch. No it is not, must be a left over from previous versions. I will remove it.

saesh:
>     * How do you handle the config.js file? It's not included in the Dockerfile (e.g. copy from template).

It is mounted as a volume to the correct location inside the Nginx container, same for `keys.js` and profiles. This works as long as the files are available on the host where the container is being run.

I personally use Ansible to copy the files to a server, and run the container on the server.

>     * Could you please add instructions how to use it to the readme, perhaps a new section "Build and run with Docker" (or so)?

Done.

README.md
```diff
@@ -44,18 +44,65 @@ As an alternative to the above online version, the standalone server of BRouter
...
+      docker build -t brouter-web .
+
+This creates a Docker image with the name `brouter-web`.
+
+#### Running Docker container
+
+To run the previously build Docker image run:
+
+      docker run --rm --name brouter-web \
+        -p 127.0.0.1:80:80 \
```
nrenner: Shouldn't this map to 8080 instead of 80, as mentioned later?
```suggestion
        -p 127.0.0.1:8080:80 \
```

saesh: Yes, that is a typo.

nrenner:
Thanks!

PR #279, opened 2020-02-25
Disable opening browser on gulp serve?

nrenner:
When running gulp serve, Browsersync automatically opens the default browser.

I personally would like to disable that, but if you want to keep it, I could probably add an env variable check.

@bagage @Phyks what is your preference?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -333,7 +333,8 @@ gulp.task(
         server.init({
             server: {
                 baseDir: './'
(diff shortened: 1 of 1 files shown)


bagage:
I have no strong opinion  on that,though I think it is valuable for newcomers. Maybe an env check is better?

Phyks:
No strong opinion for me either :)

PR #281, opened 2020-03-25
Add config option for initial map position and zoom level

rmsacks:
When I set up my own instance of brouter-web, I wanted to change the default map location closer to where I live. I had to spend a little bit of time looking through the source code to find where I could change this. I think making this a configuration option would make it easier for people to change the default map location in the future.


bagage:
Thanks @rmsacks for the improvement!

bagage:
I pushed 23b5439 to prevent current running servers to break due to outdated `config.js`.

PR #284, opened 2020-04-06
Privacy link relative path

erdmark:
The privacy link now is relative to the site that it is hosted on

When opening the about page at brouter-web the privacy link left down in the corner is a relative one to https://brouter.de/privacypolicy.html. When running an own instance I think it should link the that page, because the operator could have other terms, for example time for saving logs. So you shoud change that to a relative path to /privacypolicy.html so every operator can have another text.
You could also include a standard privacypolicy.html and put it to /brouter-web/privacypolicy.html so it can be customized if needed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -292,7 +292,7 @@
                             <a href="https://brouter.de/brouter/segments4/" target="_blank">data files</a>.
                         </p>
                         <p data-i18n="[html]about.details">
-                            <i><a href="https://brouter.de/privacypolicy.html" target="_blank">Privacy Policy</a></i
+                            <i><a href="/privacypolicy.html" target="_blank">Privacy Policy</a></i
                             >,
                             <i
                                 ><a href="https://github.com/nrenner/brouter-web#credits-and-licenses" target="_blank"


bagage:
Thanks @erdmark for this contribution! I think that indeed your PR makes sense, however I'll personally (need @nrenner thoughts on that!) prefer to have `privacypolicy.html` within the repository, because otherwise this PR would raise a `404 not found` error instead of the current redirection to brouter.de.

nrenner:
The PR only changes the default text in the about dialog, which is replaced by a translation. The translations need to be fixed on Transifex, as the languages files get overwritten on pull. Don't know if we could somehow use a placeholder for the link?

The other link is in Map.js. See https://github.com/nrenner/brouter-web/search?q=%22privacypolicy.html%22&type=Code

Not sure, but I think I prefer the suggested relative path. A default policy would need to be adjusted anyway and I'm a bit reluctant to move the site-wide brouter.de policy, mainly created by abrensch, into brouter-web. Maybe we should create a CUSTOMIZE.md with steps needed for a custom install/fork?

bagage:
Instead of modifying the translation/translated texts, maybe a simpler/better solution would be to add a config field `BR.conf.privacyPolicyUrl` which defaults to either none (then we should hide the text), or brouter.de URL to not break backward compatibility?

nrenner:
Yes, but how do you get that config variable into the text in the about dialog?

Translation texts usually allow placeholders for variable content, but I don't know if and how that would work with Transifex in this case.

bagage:
We can leave a variable within i18n text, translaters must take care of keeping the variable intact - generally, Transifex will highlight variables so it is quite easy to understand that this is a non translatable item to copy in the translated item, eg:

```html
<!-- English -->
Click here to <a href="BR.Conf.privacyUrl">show privacy policy</a>
<!-- Should/Would be translated in Transifex for French as: -->
Cliquez ici pour <a href="BR.conf.privacyUrl">afficher la politique de confidentialité</a>
```

nrenner:
But the link text is in HTML ([index.html](https://github.com/nrenner/brouter-web/blob/4bd3898f4fb1fa5f9a452c49b5332bce520ec8e7/index.html#L295)), where would that get evaluated as JavaScript?

bagage:
The HTML text is only a placeholder, as soon as l10n is loaded `data-i18n="[html]about.details"` will replace the content, doesn't it? 
I think we could actually remove all translatable content from index.html and only leave `data-i18n` attribute and it would remain the same, so only the Transifex content matters here?

nrenner:
What I mean is template-like variable replacement in translated strings, and I found in the i18next docs that they call this [interpolation](https://www.i18next.com/translation-function/interpolation).

But I haven't found on a quick search if Transifex and the [localize](https://github.com/nrenner/brouter-web/blob/4bd3898f4fb1fa5f9a452c49b5332bce520ec8e7/js/index.js#L429) function for HTML of the [jquery-i18next](https://github.com/i18next/jquery-i18next) plugin support this and I don't want to dig deeper right now.

A simpler way would be to replace it like this somewhere in our `initApp` or `initMap` functions that get called after translation:
```js
    document.querySelector("a[href='https://brouter.de/privacypolicy.html']").href = BR.conf.privacyUrl;
```

nrenner:
@erdmark Thanks! I extended your suggestion by making it configurable in e5e876f.

nrenner:
The jquery-i18next `localize` function does support interpolation by simply passing an object:

https://github.com/nrenner/brouter-web/blob/a0fe3c882496a90090d47340fe34cda1a80f99ec/js/index.js#L429-L431

So matching keys in the translation string are replaced with the value:

https://github.com/nrenner/brouter-web/blob/a0fe3c882496a90090d47340fe34cda1a80f99ec/index.html#L295

PR #290, opened 2020-05-09
Add CyclOSM in the default selection of available layers

Phyks:


nrenner:
I deliberately removed CyclOSM even as optional layer in c838b55, except for France, where its the default. This was after our mail exchange with Jocelyn (openstreetmap.fr admin) when it turned out, that our old OpenCycleMap usage (as optional layer) was 1/6th of the whole server usage.

Has anything changed since then?

Florimondable:
Hi, from my understanding of the stats https://munin.openstreetmap.fr/osm11.openstreetmap.fr/osm166.openstreetmap.fr/index.html
We're serving 1,3M tiles/day with 18% CPU time, so if your numbers were right I think the server can absorb the load.

Phyks:
Hi Norbert,

I'm sorry, I completely forgot not having it in the default selection was intentional. I've just crunched some stats.

We should be serving around 1M tiles per day currently. In total, we served 14M tiles in the last 14 days, but this is to be taken cautiously since this includes 4.4M tiles served for https://carte.velo-iledefrance.fr/ (map of covid19 temporary cycleways around Paris, launched last week).

If my stats are correct, we should serve about 60k tiles per day for BRouter currently. This is above the 18k tiles we were discussing last November, but this is expected since CyclOSM is default for French users while OpenCycleMap was featured but optional for everyone.

From these numbers, if my understanding is correct, having CyclOSM as a featured but optional layer for everyone (but kept default for French users) would represent something like 18k extra tiles per day, max.

If these numbers are correct, I think this is a reasonable increase (18k extra tiles over 1M tiles served per day in total).

If we can confirm these expected numbers, I'll loop again by email with OSM-FR before merging anything.

nrenner:
I think that we need our own rendering server. I'm just still struggling with which way I want to go for financing.

I do have more comments on numbers but need to look them up again first.

nrenner:
My proposal in November was to add it to the optional list (i.e. to the additional list behind the "More" button), to avoid too high usage and then maybe see what the actual request numbers are. The 18k requests per day were when Thunderforest layers were moved to that optional list for a short time before removal. Thunderforest layers in the default selection, like in this PR, had much higher numbers around 96k per day:

Thunderforest single tile requests around May in 2019:
|  | day | month |
|--|--|--|
| optional list |  18k | 0.6M |
| default list  |  96k | 3M |
| *2 (2020) ?   | 192k | 6M |

Our route segment request numbers have more than doubled since last year, so I accounted for that as a rough estimate in the "*2" row.

Another comparison is our openstreetmap.org tile usage that I saw recently: in the [Top 10K](https://gist.github.com/Firefishy/076cca798e0d1f947c3e5c99f1b9141c#file-top-sites-2020-02-txt-L153) websites using openstreetmap.org tiles [\[1\]](https://wiki.osmfoundation.org/w/index.php?title=Board/Minutes/2020-03-F2F/Dialogue_with_OSM_System_Administrators&oldid=7093#Current_situation) we're at rank 153: `2'354'546 brouter.de`.

Assuming this is tile requests per month in February ¹ 2020 (29 days):

|  | day | month |
|--|--|--|
| default layer ² | 81k  | 2.4M |

¹ not really outdoor season in Europe yet
² except for German and French users, which are probably our biggest user groups

Phyks:
> Thunderforest layers in the default selection, like in this PR, had much higher numbers around 96k per day:

Sure, but this includes German and French users (and French users already have CyclOSM as default).


On my side, from May, 4th to May 20th, I have:

* 1.3M tiles served from referrer brouter.de
* 72k tiles served from referrer www.brouter.de

This amounts to basically 3M tiles currently served per month for BRouter.

This is to be compared to 2.3M tiles served over the same period for cyclosm.org (our main demo website). Third referrer is [Umap](umap.openstreetmap.fr/) with 700k tiles over the same period.

Overall, we are serving around 1M tiles per day.

Given these numbers, I think we could reasonably step up a bit with CyclOSM in BRouter. If I understood correctly, current situation is CyclOSM is default for French users and in the "More" list for everyone else. We could possibly push CyclOSM to the default selection of alternative layers and see if this is manageable. This would result in a likely 20-100k extra tiles.

nrenner:
>  If I understood correctly, current situation is CyclOSM is default for French users and in the "More" list for everyone else.

No, CyclOSM is the default for French users, but not available at all for everyone else, not even in the "More" list.

This is because even with our lowest number, the 18k/day or 0.5M/month, the comment from Jocelyn was "The usage seems a bit high compared to what is currently served". The total number you had in November was 3M per month for CyclOSM. So 0.5M would have been 1/6th of that.

Florimondable:
Alright so let's go one step at a time, we could propose CyclOSM in the More list ?
(Though I don't worry much if it was in the default main list).

nrenner:
Having CyclOSM in the main list would of course be great and what you would expect from a bike-focused app. I'm very unsure what usage is still acceptable, maybe propose both options, ultimately it's on the admins to decide?

Phyks:
See https://listes.openstreetmap.fr/wws/arc/tech/2020-08/msg00000.html. This is fine for OSM-Fr to add CyclOSM in the list of layers available for everyone.

Thanks!

nrenner:
Good news, thanks!

So as discussed, we start by adding it to the list of configurable layers under "More" first?

Phyks:
Yes,I would be in favor of adopting a progressive inclusion along next two versions, if that's fine for you.

Next version: offer cyclosm under the "More" list for everyone.
Next next version: eventually move it to default set of layers for everyone, depending on monitored usage. 

Best

Bibi56:
I see "0.13.0 - No due date 100% complete" but today is 16th OpenStreetMap anniversary ;-).

nrenner:
It's already released and available as a configurable layer.

Just forgot to close the milestone. See also [changelog](https://github.com/nrenner/brouter-web/blob/master/CHANGELOG.md#brouter-web-changelog) and [releases](https://github.com/nrenner/brouter-web/releases).

Bibi56:
Perfect. Happy anniversary!

nrenner:
> Next next version: eventually move it to default set of layers for everyone, depending on monitored usage.

@Phyks now that quite some time has passed, what is the opinion on this?

Coehill:
@Phyks  @nrenner  Do we have an update on this feature?  I see that there are waymarked trails as a layer but one of the main benefits of cyclosm is that it displays bicycle infrastructure (bike lanes, separate cycleways, cycle streets, etc).

Phyks:
CyclOSM is now featured on osm.org. I expect BRouter traffic to be at most a duplicate usage of osm.org. Therefore, I see no specific reason (from CyclOSM side) to not include it in default selection. This is, of course, up to brouter-web to decide in the end given expected usage / benefit / quality of service of CyclOSM.

PR #292, opened 2020-05-13
Fix: Profile params of type 'select' ignored default value from profile

tbsmark86:
Hi,

happens on turnInstructionMode - they are always '0' regardless of setting in the profile code.


nrenner:
@Phyks want to have a look?

nrenner:
Thanks for the fix!

Issue #296, opened 2020-05-21
improve mobile responsivity


bagage:
I tweaked some files on my fork to improve web experience via PWA (progressive web app).

I'll submit a PR for that, I think we could do a TWA too quite easily so that brouterweb can be installed via app store (play store or fdroid or such). It will require #286.

I think we could improve UI in a few areas:
- [x] put everything on a single line at the bottom
- [x] put everything on a single column on the left side (or maybe remove some buttons on mobile look zoom buttons)
- [ ] improve the search icon quality that looks blurry on mobile (need to update the library but blocked by upstream issue yet https://github.com/perliedman/leaflet-control-geocoder/issues/324)
- [x] somehow improve the navbar which is not very convenient on mobile *won't do that here, it is a rather separate issue*
- [x] improve profile/alternative selection on mobile (you need double click to select an item but it seems to take effect on single click too - but related to previous point)
- [x] make modals fullscreen?
- [x] remove fullscreen button from profile layers selection on small screens
- [x] maybe more…

For now it looks like this:

![Screenshot_20200521-134854_Firefox_Preview](https://user-images.githubusercontent.com/1451988/82556706-8fd28a80-9b6a-11ea-987b-90f5f898f4ae.png)

nrenner:
Yes, going full screen would be nice.

Not quite clear to me: is the icon required for your PR or a potential future TWA?

bagage:
Not strictly required, a placeholer icon is fine too. And twa is the last item to do, once everything else is fixed.

PR #298, opened 2020-05-24
Improve mobile stats

bagage:
Related to #296, this tries to improve footer on mobile:

Capture d’écran de 2020-05-24 19-51-38

Main changes are:

- automatically fill spaces between statistics items instead of a fixed amount that does not scale well on small screens. On big screens, items are evenly distributed instead of centered. If that's an issue, I'll take a look to have some specific CSS for mobile then.
- Add a message `Please start drawing` before showing stats footer, I find the experience unpleasant at start (specifically for newcomers) - once a route has been drawn, the message disappears and stats appear. Maybe this message could provide help to users on how to get started, etc. in the future?
- Remove spaces between numbers and units, eg `4km` instead of `4 km`. Remove duplicated units on mobile to gain some space.
- Hide stats titles starting at medium screens (previously was small screens), because it wasn't fitting on a single row anyhow.
- Fixed an issue with distance title formatting, eg `3900meters` should be formatted as `3.900` but it was currently only displaying `3.9` (removing trailing zeros).

ps: trip duration seems a lot pessimistic; 80minutes for 6kms seems unbelievably slow.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -79,29 +79,52 @@ table.dataTable {
 
 footer {
     flex: none;
-
     background-color: #f7f7f9;
 }
 
+#stats-container {
+    display: flex;
+}
+
+#stats-info {
+    align-items: center;
+    height: 46px;
+    display: none;
+}
+#stats-info[style*='display: block'] {
+    display: flex !important;
+}
+#stats-info > div {
+    margin: auto;
+}
 #stats {
     flex-grow: 1;
     margin: 0;
     padding: 0;
     text-align: center;
 }
-#stats li {
-    margin: 0 1rem;
-    display: inline-block;
+
+ul#stats {
+    display: flex;
+    align-items: stretch;
+    justify-content: space-around;
+    width: 100%;
+    margin: 0;
+    padding: 0;
 }
+#stats li {
+    display: inline-block;
+    flex: 0 1 auto;
+    list-style-type: none;
+}
+
+p.stats-label {
+    margin-bottom: 0.2em;
+}
+
 @media (max-width: 767px) {
-    #stats li {
-        margin: 0 0.5rem;
-    }
     #stats {
         padding-top: 0.4em;
-    }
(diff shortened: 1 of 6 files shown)


mjaschen:
I like the improvements but would suggest to keep the space between values and units as it's typographically correct to have them separated by a space[1]. Maybe a narrow non-breaking space[2] would be a good tradeoff (that's the typographically correct version in German[3]).

[1] https://en.wikipedia.org/wiki/Space_(punctuation)#Unit_symbols_and_numbers
[2] https://en.wikipedia.org/wiki/Non-breaking_space#Width_variation
[3] https://de.wikipedia.org/wiki/Schmales_Leerzeichen#Verwendung

bagage:
Also pushed a commit to remove zoom controls on mobile, since it is quite standard to slide in/out on mobile and we need some space on left bar too. WDYT?

nrenner:
> Fixed an issue with distance title formatting, eg 3900meters should be formatted as 3.900 but it was currently only displaying 3.9 (removing trailing zeros).

I don't consider left out trailing zero decimals an actual issue, but using fixed decimals is more clear.

> Also pushed a commit to remove zoom controls on mobile, since it is quite standard to slide in/out on mobile and we need some space on left bar too. WDYT?

Ok with me, no strong opinion here.

bagage:
> I like the improvements but would suggest to keep the space between values and units as it's typographically correct to have them separated by a space[1]. Maybe a narrow non-breaking space[2] would be a good tradeoff (that's the typographically correct version in German[3]).
> 
> [1] https://en.wikipedia.org/wiki/Space_(punctuation)#Unit_symbols_and_numbers
> [2] https://en.wikipedia.org/wiki/Non-breaking_space#Width_variation
> [3] https://de.wikipedia.org/wiki/Schmales_Leerzeichen#Verwendung

Yeah I'll fix that in a separate PR.

PR #303, opened 2020-06-06
Allow to mute route display

stesie:
Having used Komoot before, I really like their feature to "mute" the route display -- especially since I configured it to full opacity and different color (I struggle to see the slightly transparent magenta line well).

So this PR adds the ability to temporarily mute/hide the route, by holding down the "m" key.

I've added it to the `OpacitySlider` input, even so it's not really a feature of the slider itself. However it's currently the only place to track the current opacity value and I didn't want to scatter responsibilities there. Let me know if you see things differently


nrenner:
Fine with me, thanks!

Issue #249, opened 2019-10-03
Using ES 6 or parts of it?


nrenner:
We currently don't use ES 6.

My idea was to either wait until we use Babel (with Webpack?) or to properly define what Browsers we want/need to support and what features we can safely use, ideally by looking at user agent strings from user requests.

The first thing would be to check the code with eslint-plugin-compat (#188) against our current browserlist (which is Browsers supporting `flexbox`):

https://github.com/nrenner/brouter-web/blob/029bdad4ff95f483119e936804b9791a49cc2c28/package.json#L25

PR #304, opened 2020-06-07
Add sidebar tab with route analysis data

mjaschen:
- adds a new analysis tab in the sidebar
- the analysis tab shows length distribution for different way types, surface, and smoothness
- table rows can be hovered/clicked to highlight the according segments on the map (similar behaviour as the detailed data table)
- localization is implemented for `de` and `en`
- the method for finding segment edges was extracted from `js/control/TrackMessages.js` into `js/util/TrackEdges.js` as it's used in the new analysis class too (the Gulp config was changed to reflect that)

Notes:

I had the idea to use the *DataTable* plugin for rendering the tables but decided against it. The only meaningful way to sort such a table is by the length column and that's already the case. So it's just three plain, old HTML tables, rendered by jQuery.

For meaningful statistics the `processUnusedTags` setting has to be enabled in the routing profile. Only in this case the BRouter backend includes all needed tags (`highway`, `surface`, and `smoothness`) for *every* route segment in the response. I’ve enabled that setting for all profiles at my BRouter-web instance at <https://brouter.m11n.de/>.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -249,6 +249,38 @@ input#trackname:focus:invalid {
     margin: 0;
 }
 
+.track-analysis-header-distance {
+    text-align: right;
+}
+
+.track-analysis-table td {
+    font-size: small;
+}
+
+table.dataTable.track-analysis-table tfoot td {
+    font-weight: bold;
+    text-align: right;
+    padding-right: 0;
+}
+
+table.track-analysis-table tr:hover {
+    background-color: rgba(255, 255, 0, 0.3);
+}
+
+.track-analysis-title {
+    text-transform: capitalize;
+}
+
+.track-analysis-distance {
+    text-align: right;
+}
+
+.track-analysis-heading {
+    margin-top: 15px;
+    font-weight: bold;
+    font-size: 1.2em;
+}
+
 /* dashed line animation, derived from Chris Coyier and others
 https://css-tricks.com/svg-line-animation-works/
 */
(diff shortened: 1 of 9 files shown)


bagage:
A side note on translation: we do not support adding translations directly yet (eg `de.json` modifications) since that will be overwritten when pulling from transifex.
If many translations were done, we can upload the json to transifex but casually to not erase remote translations @nrenner.

mjaschen:
> A side note on translation: we do not support adding translations directly yet

Is the translation workflow somewhere documented?

bagage:
Unfortunately no, I'll begin a CONTRIBUTING.md to add notes about that!

nrenner:
Thanks for your contribution, great addition!

I entered the German translations manually in Transifex. For the tag value translations see #306.

nrenner:
Regarding `processUnusedTags` I opened an issue:
[Process additional tags for analysis · #249 · abrensch/brouter](https://github.com/abrensch/brouter/issues/249)

nrenner:
I committed a fix for the sort compare function in 71317b8.

The return value of the [sort compare function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) is defined as value less, equal or greater than 0, not as boolean.

For some reason sorting worked in Firefox anyway, but not in Chromium.

mjaschen:
Thanks!

PR #308, opened 2020-06-08
add docker-compose tasks for development

schmic:
I usually try to avoid installing tools for development processes.

That PR enables everyone to use docker(-compose) instead of manually installing node/npm and yarn.


nrenner:
I've heard from people without Node/Yarn experience having difficulties getting started, so sure will be helpful for those also.

Thanks for your contributions!

PR #314, opened 2020-06-15
Add more keyboard shortcuts and fix various shortcut related issues

rkflx:
Currently there are only a couple of keyboard shortcuts available:
- +/- for zooming, arrows for panning
- D for drawing
- P for POIs
- M for muting

Let's add more shortcuts, to make BRouter-Web more efficient to use:
- L for geolocation
- F for finding
- R for route reversing
- N for no-go areas
- Z for undoing
- Backspace for clearing
- C for color coding
- S for the Strava layer
- E for toggling the elevation chart
- G for changing profile
- X for Export
- O for opening tracks, Shift+O for opening no-go areas
- H for the helpful about dialog
- T to toggle sidebar tabs, Shift+T to switch tabs

With this, the mouse pointer can be kept in the center of the screen to manipulate the map, without having to travel a large distance to the side to click on tiny buttons, making it easier to quickly switch between various functions and use BRouter-Web to its full potential.

While at it, fix a couple of bugs and make the behavior more standard:
- Stop deleting points once there are no more points left
- Allow translation of the search field placeholder text
- Fix mute shortcut not working for color coded routes
- Prevent browser keyboard shortcuts from triggering route functions
- Prevent triggering shortcuts from modal dialogs
- Prevent triggering shortcuts from open dropdown menus
- Inhibit shortcuts in number input fields too
- Add missing shortcut tooltips to control buttons
- Convert all shortcuts to act on keydown already


rkflx:
(Had to update 02af638, see comment there. Any hints? At least the geocoder shows results again, but the fact I have to use `setOptions` instead of directly extending `options` above is a bit fishy in the first place...)

nrenner:
So just adding it to the [options](https://github.com/nrenner/brouter-web/blob/8621c4403b642a1e744a6d93ee947033563d768a/js/plugin/Search.js#L2) list (without an `initialize` method) didn't work?

rkflx:
> So just adding it to the [options](https://github.com/nrenner/brouter-web/blob/8621c4403b642a1e744a6d93ee947033563d768a/js/plugin/Search.js#L2) list (without an `initialize` method) didn't work?

Well, to be more specific: IIRC it did work from a Leaflet point of view (e.g. with a plain string), but `i18next.t` would only return `undefined` at this point in time.

However, we need some kind of initializer to execute the code which binds to `_keydownListener` (776730a) anyway, i.e. we have to work around two issues at the same time here.

I'll add one more comment to the code...

nrenner:
I see that `i18next.t` can't work in a static context as the options declaration.

Haven't seen the old commit, did you call super `initialize` (needs to be done when overriding it)? This seems to work for me:
```js
    initialize: function(options) {
        L.Control.Geocoder.prototype.initialize.call(this, options);

        L.setOptions(this, {
            placeholder: i18next.t('map.geocoder-placeholder', 'Suchen...')
        });
    },
```

rkflx:
Thanks for the hint, which does the trick indeed. Apparently I only looked at examples in the codebase where this was not needed, but now I see this being done elsewhere too.

> Haven't seen the old commit

In general, clicking on "force-pushed" should show everything that changed.

rkflx:
Rebase

Update shortcut inhibitions:
- Prevent triggering shortcuts in number inputs too
- Prevent triggering shortcuts in dropdowns too

nrenner:
Thanks for your thorough work!

nrenner:
I'm holding off pushing to Transifex for now, maybe we can make translator's life easier: #315?

PR #317, opened 2020-06-20
Combine and reuse translations for keyboard shortcuts

### LINKED ISSUE #315: Combine and reuse translations?
#314 appends keyboard shortcuts to tooltip texts like this (see [en.json diff](https://github.com/nrenner/brouter-web/pull/314/files#diff-772a593ff61f0e484d43cc349a5ab31cR129)):

```json
"reverse-route": "Reverse route (R key)",
```

I wonder if we can make translator's life easier by combining the tooltip text itself and the key appendix in a more generic fashion.

Maybe by using [nesting](https://www.i18next.com/translation-function/nesting), leaving the original translations untouched and using a new translation key that concatenates both, e.g. (untested):

```json
  "keyboard": {
    "shortcut": "key"
  },
  "map": {
    "reverse-route": "Reverse route",
    "reverse-route-key": "$t(map.reverse-route) (R $t(keyboard.shortcut))"
```

Or, maybe if [append](https://github.com/i18next/jquery-i18next#append-content) works with attributes like this (also untested):
```html
data-i18n="[title]map.reverse-route;[title][append]keyboard.shortcut"
data-i18n-options='{ "key": "R" }'
```
with
```json
  "keyboard": {
    "shortcut": " ({{key}} key)"
  },
```

rkflx:
Using i18next's "nesting" and "interpolation" features should reduce the overall number of strings to translate, as well as provide a standardized pattern for shortcut tooltips (if applicable).

Note that this approach is still allowing for flexibility regarding differently structured sentences in each language.

Resolves #315

Bonus commits:
- Run `yarn lint`, `yarn prettier` and `yarn build i18next`
- Add warning regarding `yarn build i18next` / `gulp i18next`
- Switch to next tab with Shift+T shortcut even if tab is currently closed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,9 +48,9 @@ docker-compose up serve
 
 ### How internationalization works
 
-BRouter is translated using [i18next](https://www.i18next.com/) library, via command `gulp i18next`. It extracts translatable elements into `locales/en.json` file (English version).
+BRouter is translated using [i18next](https://www.i18next.com/) library, via command `gulp i18next`. It extracts translatable elements into `locales/en.json` file (English version). (Note that unused translation keys or keys not referenced in `keys.js` might get removed automatically. Make sure to commit any changes first before running this, and only amend the previous commit after checking the diff carefully.)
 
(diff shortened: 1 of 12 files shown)


nrenner:
Thanks!

PR #328, opened 2020-08-02
Usability fix: Allow Return Key in export dialog

tbsmark86:
very simple but anoying for a keyboard user :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -544,7 +544,7 @@
                                     Close
                                 </button>
                                 <button
-                                    type="button"
+                                    type="submit"
                                     class="btn btn-primary"
                                     data-i18n="export.title"
                                     id="submitExport"


bagage:
Thanks :)

rkflx:
This was one of the annoying things I did not yet get around to tackle when working on #314. Thanks for trying to fix this!

However, it does not work in Chromium for me, where it even breaks for mouse users: Starting at an URL with an empty query component, clicking "Export route" in the export dialog redirects to `/?format=gpx#...`.

(In Firefox the issue does not manifest itself, since there the `/?format=gpx` request is listed as "blocked".)

nrenner:
I added a `event.preventDefault()` to stop the default submit action and it seems to fix the download and `?format=gpx` URL query in Chromium.

Don't know why the return key does not work in Chromium. From a quick search the criteria for implicit submission is a `form` tag and non-submit buttons having a `type="button"` attribute which both seems to be the case.

rkflx:
@nrenner Thanks, working great now.

PR #341, opened 2020-10-09
Fix color coding by cost. Closes #340.

### LINKED ISSUE #340: Cost color coding uses the total cost instead of the cost per distance
I discovered today that it is possible to visualize the cost for each section using color-coding. That is fantastic, thank you so much for that!

However, I have discovered something curious: the color seems be computed based on the total cost for each section. I would expect it to be computed based on the cost per distance instead.

Consider the following example:
image

https://brouter.de/brouter-web/#map=11/47.4515/6.2203/cyclosm,route-quality&lonlats=6.134818,47.398349;6.435182,47.553823&profile=vm-forum-velomobil-schnell

The long red segment looks really scary, right? But in fact it is just because this is the longest segment of the path (14 km). The elevation profile is pretty regular, this section is paved like others, so there is no reason to display this as particularly hard.

I would instead expect the color to be computed using the "$/km$ variable exposed by the statistics tab.

wetneb:
This fixes color coding by cost (#340), making sure it is relative to the length of the path.

Two points where I am not quite sure:
* I used the same technique as the `incline` provider to compute the distance, but this is probably inaccurate: instead of the distance between the start and the end point, the length of the path should probably be used. I suspect this is available in the `feature.cost` object, no?
* It is not clear to me whether the `initial` and `node` costs should also be divided by the distance (they were all 0 in my example). I am sure the `elev` and `turn` costs should be divided since they seem to be integrated over the whole segment.

### DIFF
--- a/js/plugin/RoutingPathQuality.js
+++ b/js/plugin/RoutingPathQuality.js
@@ -68,16 +68,15 @@ BR.RoutingPathQuality = L.Control.extend({
                         outlineColor: 'dimgray',
                         renderer: renderer
                     },
-                    valueFunction: function(latLng, prevLatLng) {
+                    valueFunction: function(latLng) {
                         var feature = latLng.feature;
-                        var distance = Math.min(prevLatLng.distanceTo(latLng), 1) / 1000; // in km
-                        return (
-                            feature.cost.perKm +
-                            (feature.cost.elev +
-                             feature.cost.turn +
-                             feature.cost.node +
-                             feature.cost.initial) / distance
-                        );
+                        var cost = feature.cost.perKm;
+                        if (feature.distance > 0) {
+                            cost +=
+                                (feature.cost.elev + feature.cost.turn + feature.cost.node + feature.cost.initial) /
+                                feature.distance;
+                        }
+                        return cost;
                     }
                 })
             }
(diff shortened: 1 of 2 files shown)


nrenner:
Thanks for looking into this, I hadn't noticed.

I think the distance of the whole path/edge should be used instead of the single coordinate distances.

That distance is returned from the server in the `messages` property but not passed in the `feature` object yet. It probably could be added in [`Brouter._getFeature`](https://github.com/nrenner/brouter-web/blob/652de90c06aa8a696a2fedce316ddc686d01bd15/js/router/BRouter.js#L230).

wetneb:
Makes sense, here is a fixed up version.

That being said, the visualization is still disappointing even with that fix, because the elevation cost is computed for the entire path. Instead, we would need to break down the elevation cost returned by the server for each segment in the path, as the `incline` provider does. I have no idea if this is something the server returns at all (or even if it makes sense - the elevation cost might not be easily broken down into each segment).

nrenner:
Thanks for your contribution!

It wouldn't make sense to break down the cost for each line between two coordinates as the incline provider does. The router can only decide whether or not to take an edge between two junction nodes and therefore needs to look at the whole cost for the entire edge - and the cost coloring should reflect that.

The problem in this case is, that subsequent edges are [aggregated](https://github.com/abrensch/brouter/blob/0841216cda47d3327b039e6db00c253e86d0948d/brouter-core/src/main/java/btools/router/OsmTrack.java#L127-L155) in the messages array when the tags are the same. Which is a bit unfortunate when analyzing elevation costs. I opened a separate issue for this: #344.

Issue #343, opened 2020-10-17
Upgrade to the latest version of d3 5.x


alexcojocaru:
Using an older version of d3 (i.e. 3.x) in the brouter-web app makes it difficult to integrate leaflet plugins which depend on the latest d3 (i.e. 5.x atm). When trying to integrate the [Leaflet.Heightgraph](https://github.com/GIScience/Leaflet.Heightgraph), as a replacement for the Leaflet.Elevation currently used to render the elevation profile of the route, and since Heightgraph depends on d3 5.x, I only had two options:

1. Backport the Heightgraph to d3 3.x, which I have done [without too much trouble](https://github.com/GIScience/Leaflet.Heightgraph/compare/master...alexcojocaru:d3-v3_2); going forward with this option means that:
   1. someone would have to maintain the fork and release it whenever needed (https://github.com/GIScience/Leaflet.Heightgraph/issues/101)
   1. as more and more d3 5.x APIs, with no direct translation to d3 3.x APIs, are used by the Heightgraph (e.g. [the color category](https://github.com/GIScience/Leaflet.Heightgraph/compare/master...alexcojocaru:d3-v3_2#diff-c50ee78a6e8e50ed26533866c2055c843c0fbaa273a90baf8399b96e217946f5R330)), the backporting could get complex and the fork could diverge significantly
1. Upgrade d3 to the latest 5.x in the brouter-web, so we could integrate the Heightgraph as it; an integration plugin (similar to the [Elevation integration plugin](https://github.com/nrenner/brouter-web/blob/master/js/plugin/Elevation.js)) would be required, but I have written it already - it is simpler than the Elevation integration, thanks to the resize support in Heightgraph, as well as other APIs which I have already integrated into the project to make it easier to embed.
This was my first choice and I took a stab at it, but it got pretty hairy and I fell back to the other option (backporting the Heightgraph to d3 3.x). FWIW this would be the cleanest option, and @nrenner suggested that I should start a discussion on this, so here we go...

On a side note, here's a screenshot of a route in brouter-web with the Heightgraph integrated and enabled:
![screenshot](https://user-images.githubusercontent.com/479131/96327306-b69bcc80-0ff5-11eb-9aa5-e691adc1faf9.png)

nrenner:
The reason I wanted to discuss is to see if there is a way forward with option 2, as I am planning to add ES6 syntax support through Babel anyway - but no modules yet. 

[...]

Issue #203, opened 2019-06-04
No calculation of distance, travel time, etc.


Umea123:
No calculation of distance, travel time, etc. in Northern Europe / north of the 60 ° latitude.

nrenner:
Thanks for reporting.

Example:
http://brouter.de/brouter-web/index.html#map=8/70.813/24.620/osm-mapnik-german_style&lonlats=25.740967,71.13454;23.642578,70.678153

This probably is because of no [elevation data](http://srtm.csi.cgiar.org/srtmdata/) beyond 60 degrees north and south and the recent change to remove elevation points with no data (#147).

When no elevation points remain, the Elevation plugin throws an exception because of empty data:

```
BRouter.js:188 Error: no data in parameters
    at e._calculateFullExtent (L.Control.Elevation.js:648)
    at e._applyData (L.Control.Elevation.js:745)
    at e.addData (L.Control.Elevation.js:668)
    at e.update (Elevation.js:80)
    at n (index.js:245)
    at o (L.Routing.js:340)
    at e.<anonymous> (Routing.js:317)
    at Object.callback (L.Routing.js:396)
    at r (async.js:798)
    at async.js:32 "..."
```

[opoto/Leaflet.Elevation - src/L.Control.Elevation.js#L645-L649](https://github.com/opoto/Leaflet.Elevation/blob/d3a7bcbef62261138417792900d63c3091c73dbe/src/L.Control.Elevation.js#L645-L649):
```js
    _calculateFullExtent: function(data) {


        if (!data || data.length < 1) {
            throw new Error("no data in parameters");
        }
```

Which prevents the subsequent stats update code to be executed.

With the older version at damsy.net there are some "NaN" errors (Chrome) / warnings (Firefox) in the console, but don't seem to matter:
https://brouter.damsy.net/latest/#map=8/70.813/24.620/osm-mapnik-german_style&lonlats=25.740967,71.13454;23.642578,70.678153

nrenner:
There is another smaller issue, that the beginning of a route with missing data is not shown in the diagram, not really wrong but misleading:
http://brouter.de/brouter-web/#map=12/59.9891/5.5060/Stamen.Terrain&lonlats=5.51033,60.028328;5.506211,59.9679

[...]

PR #345, opened 2020-10-27
Use Heightgraph in lieu of Elevation to show the elevation profile

alexcojocaru:
Things to note:
* d3 3.x was used only by Elevation, I removed both from the project dependencies;
  that is because Heightgraph only works with d3 5.x, whereas this project
  used d3 3.x;
* Since Heightgraph is written with ES6 (as is d3 5.x I believe),
  I had to use its transpiled bundle (which includes d3 5.x);
* There were a few definitions in the transpiled Heightgraph which conflicted
  with the rest of brouter-web; I had to comment them out via a patch
  applied in the postinstall phase.

This PR is related to https://github.com/nrenner/brouter-web/issues/343;
since I used the transpiled Heightgraph, the upgrade to d3 5.x is not necessary.


nrenner (changes requested):
Thanks for working on this!

Some issues:
1. Leaflet.Heightgraph and the plugin wrapper Heightgraph.js don't seem to handle missing elevation data:
   - `TypeError: Cannot read property 'text' of  undefined  at NewClass._prepareData (L.Control. Heightgraph.js:4994)`
     - http://localhost:3000/index.html#map=12/53.5235/8.5705/osm-mapnik-german_style&lonlats=8.545475,53.520411;8.565903,53.529595;8.584442,53.538267&profile=fastbike
   - `TypeError: Cannot read property 'geometry' of  undefined  at Heightgraph.js:267`
     - http://localhost:3000/index.html#map=17/44.44707/15.18322/standard&lonlats=15.183293,44.447082;15.181603,44.448656&profile=fastbike
     - http://localhost:3000/index.html#map=12/59.9929/5.4975/Stamen.Terrain&lonlats=5.51033,60.028328;5.506211,59.9679
   - see issues abrensch/brouter#137, #203 and  [Leaflet.Elevation#63 (comment)](https://github.com/MrMufflon/Leaflet.Elevation/issues/63#issuecomment-581027804)
2. collapsed state is not remembered for the next session 
   - this is due to missing `initCollapse` method and it's call in `index.js`
3. missing bottom tiles after resizing window and closing height graph
   - `map._onResize();` call after collapse missing, also in `initCollapse` method
4. applying a patch to a distribution bundle is a bit fragile and will probably fail (need maintenance) with some future update
   - wrapping the bundle inside a closure using an  IIFE (immediately invoked function expression),  so d3 doesn't pollute the global scope, seems to  work as well:
     ```
     (function() {
       ...
     }());
     ```
     Automatically adding those lines at the top and bottom of the bundle file would be less fragile, so I would prefer that instead of the patch, if there is a way

alexcojocaru:
Thanks for looking at this, @nrenner . I will have a look at issues 1 through 3 and fix them.

About 4 (patching the Heightgraph bundle): 
* I set a fixed Heightgraph version in package.json, which shouldn't break the patch unless the Heightgraph dependency is explicitly updated, but I agree that this patch is fragile and it would probably have to be redone when switching to newer versions of the dependency, so it's just a temporary solution
* there's [this discussion](https://github.com/GIScience/Leaflet.Heightgraph/issues/101) which I started on the Heightgraph project, where I asked if it's possible to wrap the whole code in the bundle; given there hasn't been much feedback, I expect the worst (i.e. not going to happen any time soon, if any)
* do you know if it's possible to modify the patch to apply the IIFE? that would make it more resilient to further changes to the Heightgraph bundle; if so, wouldn't there be issues with extending Heightgraph [here](https://github.com/nrenner/brouter-web/pull/345/files#diff-781f92da3be7c52bcdf8bd0d836ae292a1f35798a78b24309d5290706f6c3adcR2) ?

nrenner:
> it looks like I can modify the patch to apply the IIFE [...] does this sound like an acceptable solution for the time being?

I guess that would already be an improvement and maybe easier to check if the patch is still ok.

I intend to open separate/new Heightgraph issues for 1. and 4.

nrenner:
What I have in mind is the same approach I implemented for Leaflet.Elevation in PR MrMufflon/Leaflet.Elevation#84 (we're using my fork).

Leaflet.Heightgraph should support `undefined`/`null` values for all or part of the elevation values and show them as gap in the graph using a d3 [defined](https://github.com/d3/d3-shape/blob/master/README.md#area_defined) accessor function.

nrenner:
I opened an issue for 4.:  GIScience/Leaflet.Heightgraph#107.

The minified bundle has an IIFE, so we might just use that for now, but for some reason it didn't get included when I had a quick try changing the main reference to it.

alexcojocaru:
> The minified bundle has an IIFE, so we might just use that for now, but for some reason it didn't get included when I had a quick try changing the main reference to it.
It's not obvious why, all I can say is that the Heightgraph attribute is not available in that scope.

I have been working on handling points with `undefined` altitude coordinate without errorring out, as there is some manipulation of such points in the plugin. Once that is done, I will look at issues #2 and #3 you described earlier.

alexcojocaru:
A short update on the progress:
1. Missing elevation data: the code in Heightgraph.js can handle such points without breaking; I have tested many edge cases and it works as expected. Leaflet.Heightgraph has to be fixed to deal with such point on its end though.
2-3. My bad on ignoring some of the code in Elevation.js (resize, collapse/expand, etc). I worked on an old version of your project and did not notice these improvement, nor they conflicted with my changes (since Elevation.js is removed in my branch).
4. The patch was updated to use an IIFE. Still couldn't get the minified Heightgraph to work.

alexcojocaru:
Just finished fixing items 2 and 3.

To summarize: the only outstanding issue currently is in Leaflet.Heightgraph, when the track contains points without elevation data. Given the recent activity on that project (or the lack thereof), I am not too optimistic about it being fixed soon.

alexcojocaru:
Like I said in the previous comment, I have the feeling that the bug in Leaflet.Heightgraph around handling points without elevation would take quite a while to get fixed. Because of this, I handled such points in the Heightgraph plugin and inferred their elevation from the closest points with elevation in the track: https://github.com/nrenner/brouter-web/pull/345/commits/d83fddd4edcbf3abcfb68d822102549099fa0f43

Speaking of this bug in Leaflet.Heightgraph, since you fixed the similar one in Leaflet.Elevation and you have more context, could you open an issue for it please? Thanks.

nrenner:
Thanks for the updates.

I'm not sure about interpolating missing elevation values. While interpolation might work for some cases, it will be misleading or just wrong for others. I personally consider showing the data as it is - with gaps in the graph for missing values - as the proper solution, which is why I implemented this for Leaflet.Elevation. Switching to Heightgraph without this feature feels like a step back in this regard.

I have opened an issue for missing elevation data: [Heightgraph#108](https://github.com/GIScience/Leaflet.Heightgraph/issues/108). And I will look into making a PR for this.

alexcojocaru:
👍  to your statement. Undoing it is easy, I just have to revert the last commit.
Looking forward to having the Heightgraph issue fixed...

nrenner (approved):
Thanks for your contribution!

alexcojocaru:
Thanks for integrating this change, I look forward to using it.

My impression was that the major blocker to using Heightgraph was its inability to handle tracks with points with undefined elevation. I see that you [fixed it](https://github.com/GIScience/Leaflet.Heightgraph/pull/110), and the IIFE issue as well, but the PRs were not merged in.

PR #347, opened 2020-11-29
Add 20km allowed zone

bagage:
Due to the pandemic, in France at least, rides are limited to a 20km circle-area around our home.

I've added it on my instance - do you think it would be useful to have it on main instance as well? There might be some work needed however, for instance hiding the icon for users not impacted, and maybe make the radius configurable if the 20km limit is different in other countries?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -197,6 +197,9 @@ input#trackname:focus:invalid {
 .pois-draw-enabled {
     cursor: cell;
 }
+.circlego-draw-enabled {
+    cursor: pointer;
+}
 
 #map {
     /* center error message horizontally */
--- a/index.html
+++ b/index.html
@@ -373,6 +373,7 @@
                                     Source
                                 </legend>
                                 <p>
+                                    <input type="text" id="nogoJSON" hidden />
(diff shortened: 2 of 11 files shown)


nrenner:
Yes, I guess it might be a useful addition for affected users.

Ireland seems to have a 5 km radius at the highest level, but not currently. Don't know about other countries.

So I think I would prefer to show that button in France only:
- a simple workaround might be to check for `fr` in the first entry of `i18next.languages`, e.g. see [LayersConfig.js](https://github.com/nrenner/brouter-web/blob/1958f55f18aec00729cb6de56f10545ddfe65c85/js/LayersConfig.js#L98)
- or maybe add a simplified boundary GeoJSON and do a point-in-polygon check using Turf.js with the map center

bagage:
Thanks @nrenner! 

I've finalized the PR and used the `i18next.languages` trick to detect country code (thanks for the tip). 
Using a GeoJSON boundary would be nice but since this is a 2 weeks feature (hopefully!), the i18next solution seems good enough to me.

bagage:
ps: if merged, a `yarn push-transifex` should be done since I added some strings (including extra error messages).

PR #350, opened 2020-12-02
Fix limit area issues

### LINKED ISSUE #349: 20km allowed zone issues
I just actually tested the PR #347 and have the following issues:

1. Error when `circleGo` not added (language not `fr`)
   https://brouter.damsy.net/latest/?lng=en
   ```
   CircleGoArea.js:134 Uncaught TypeError: Cannot read property 'getLayers' of null
       at e.getCircle (CircleGoArea.js:134)
   ```
2. When removing the zone, all no-go areas are also deleted:
   - https://github.com/nrenner/brouter-web/blob/779c720b7d55574266fb61787033fb41c8e3d0c0/js/plugin/CircleGoArea.js#L93
3. Sometimes it does find a way out, maybe polyline is not a closed ring?
     - [https://brouter.damsy.net/latest/?lng=fr#map=11/48.8855/2.3720/standard&lonlats=2.136841,48.942799;2.043457,48.931972&circlego=2.361374,48.940091,20000&polylines=2.361374,49.119754, ...](https://brouter.damsy.net/latest/?lng=fr#map=11/48.8855/2.3720/standard&lonlats=2.136841,48.942799;2.043457,48.931972&circlego=2.361374,48.940091,20000&polylines=2.361374,49.119754,2.334468,49.118886,2.307823,49.11629,2.2817,49.11199,2.256352,49.10603,2.232025,49.098468,2.208956,49.089375,2.187369,49.078842,2.167471,49.06697,2.149454,49.053874,2.133493,49.039682,2.119739,49.024532,2.108324,49.008569,2.099356,48.991948,2.092919,48.97483,2.089073,48.957381,2.087852,48.939768,2.089265,48.922161,2.093296,48.90473,2.099903,48.887642,2.10902,48.871061,2.120557,48.855147,2.134402,48.840053,2.15042,48.825921,2.168455,48.812889,2.188334,48.80108,2.209866,48.790607,2.232844,48.781571,2.257048,48.774057,2.282247,48.768137,2.3082,48.763868,2.33466,48.76129,2.361374,48.760428,2.388088,48.76129,2.414548,48.763868,2.440501,48.768137,2.4657,48.774057,2.489904,48.781571,2.512882,48.790607,2.534414,48.80108,2.554293,48.812889,2.572328,48.825921,2.588346,48.840053,2.602191,48.855147,2.613728,48.871061,2.622845,48.887642,2.629452,48.90473,2.633483,48.922161,2.634896,48.939768,2.633675,48.957381,2.629829,48.97483,2.623392,48.991948,2.614424,49.008569,2.603009,49.024532,2.589255,49.039682,2.573294,49.053874,2.555277,49.06697,2.535379,49.078842,2.513792,49.089375,2.490723,49.098468,2.466396,49.10603,2.441048,49.11199,2.414925,49.11629,2.38828,49.118886,2.361374,49.119754&profile=fastbike)
   - alternative might be to not use a no-go area, just add the polygon as a layer to the map for visual reference and maybe check with Turf.js if the resulting track is fully inside the polygon and show a warning?
     - or use an [inverted polygon](https://gist.github.com/nrenner/0de6cea585216540b6ed90a730a76134) like this (world bounds as outer and zone as inner hole) and check for intersection with track

bagage:
> 1. Error when circleGo not added (language not fr)

Indeed, I massively modified the code and forgot to retest that at the end, my bad.

> 2. When removing the zone, all no-go areas are also deleted: 

Yes, I couldn't find an easy way to only remove the circleGo area since I don't know which one it is. I pushed a minor to fix to avoid erasing nogo areas on URL load, but once you set/move the home icon, it'll still remove nogo areas. Do you see any workaround here?

> 3. Sometimes it does find a way out, maybe polyline is not a closed ring? 

Nice catch :). The ring should be closed, not sure what's happening here; I reduced the ring segments from 64 to 32 (still looks OK) and the issue went away.. maybe 64 points is too much for turf or brouter engine? 
 
Fix #349

### DIFF
--- a/js/control/Message.js
+++ b/js/control/Message.js
@@ -50,6 +50,8 @@ BR.Message = L.Class.extend({
             err = i18next.t('warning.invalid-route-to');
         } else if (err == 'from-position not mapped in existing datafile\n') {
             err = i18next.t('warning.invalid-route-from');
+        } else if (err && err.startsWith('null description for: ')) {
+            err = i18next.t('warning.no-route-found');
         }
         this._show(err, 'error');
     },
--- a/js/index.js
+++ b/js/index.js
@@ -34,7 +34,7 @@
             drawButton,
             deleteRouteButton,
             pois,
-            circleGo,
+            circlego,
             urlHash;
 
         // By default bootstrap-select use glyphicons
@@ -260,8 +260,6 @@
         });
 
         pois = new BR.PoiMarkers(routing);
-        circleGo = new BR.CircleGoArea(routing, nogos, pois);
-        pois.circlego = circleGo;
 
         exportRoute = new BR.Export(router, pois);
 
@@ -322,18 +320,22 @@
 
         nogos.addTo(map);
 
-        var shouldAddCircleGo = false;
+        var circlegoRadius = null;
         var lang = i18next.languages.length && i18next.languages[0];
 
         if (lang.startsWith('fr')) {
-            circleGo.options.radius = 20000;
-            shouldAddCircleGo = true;
+            circlegoRadius = 20000;
         }
 
-        if (shouldAddCircleGo) circleGo.addTo(map);
+        if (circlegoRadius != null) {
+            circlego = new BR.CircleGoArea(routing, nogos, pois);
+            circlego.options.radius = circlegoRadius;
+            pois.circlego = circlego;
+            circlego.addTo(map);
+        }
 
         var buttons = [drawButton, reverseRouteButton, nogos.getButton()];
-        if (shouldAddCircleGo) buttons.push(circleGo.getButton());
+        if (circlegoRadius) buttons.push(circlego.getButton());
         buttons.push(deletePointButton, deleteRouteButton);
 
         L.easyBar(buttons).addTo(map);
@@ -392,10 +394,10 @@
             var opts = router.parseUrlParams(url2params(url));
             router.setOptions(opts);
             routingOptions.setOptions(opts);
-            if (opts.circlego) {
+            if (circlego && opts.circlego) {
                 // must be done before nogos!
-                circleGo.options.radius = opts.circlego[2];
-                circleGo.setCircle([opts.circlego[0], opts.circlego[1]]);
+                circlego.options.radius = opts.circlego[2];
+                circlego.setCircle([opts.circlego[0], opts.circlego[1]]);
             }
             nogos.setOptions(opts);
             profile.update(opts);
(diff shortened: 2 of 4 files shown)


nrenner:
> I couldn't find an easy way to only remove the circleGo area since I don't know which one it is. 

I haven't thought this through, but instead of using `NogoAreas.uploadNogos()`/`.clear()` maybe `NogoAreas` could be extended with two methods for adding and removing a single no-go layer (and firing update) to/from the [drawnItems](https://github.com/nrenner/brouter-web/blob/779c720b7d55574266fb61787033fb41c8e3d0c0/js/plugin/NogoAreas.js#L47) `FeatureGroup` where no-go areas are stored. `CircleGoArea` would create that layer and keep a reference to it.

As mentioned in 3. another alternative would be not use no-go areas at all for this, but a client-only approach to maintain a separate layer on the map and check each route update with Turf if the track is outside the area and show a warning.

But maybe using both features together is rare and as this is useful now and hopefully needed for a limited time only, I guess we can also live with it.

> I reduced the ring segments from 64 to 32 (still looks OK) and the issue went away

Not for me, it even finds a shorter way through. Moving the zone around, keeping the start inside and end outside, it seems like it finds a route through about half of the time.

Probably a flaw in the BRouter algorithm. Maybe using two half lines instead of a single ring might be worth trying out.

nrenner:
> As mentioned in 3. another alternative would be not use no-go areas at all for this

The advantage of the no-go ring is that it helps with finding routes that stay inside.

Let me know if this is still a WIP, I wouldn't mind merging as is.

bagage:
Yes the turf client-side only feature would be a great additional enhancement - but I won't have time for it right now. If this feature is kept permanently later (hopefully no!), this could be added.

Modifying Nogo could do the trick but again, I quickly looked at it and I thought it would need several more hours that I cannot spend on it right now. Same as previous comment, I'd rework on it if the feature lasts more than 1/2 months.

I'll only test to split the ring into 2 chunks to test if it fixes the problem - if not, let's merge it as is. I'll ping you when it's ready :).

bagage:
OK so here is another attempt, using 2 half rings as suggested @nrenner. If it fixes the issue for you too, then it is ready to be merged as-is :). I'll try a bit more to reproduce the routing issue with this version but for now it seems good.

nrenner:
> OK so here is another attempt, using 2 half rings

That seems to fix it, thanks!

> As mentioned in 3. another alternative would be not use no-go areas at all for this
>> The advantage of the no-go ring is that it helps with finding routes that stay inside.
>>> Yes the turf client-side only feature would be a great additional enhancement

The client-side Turf check actually wasn't such a great idea. Here is an example where the route would take the bridge outside and a check can just show a warning/error, whereas with the no-go ring the router finds an alternative route inside:

http://localhost:3000/?lng=fr#map=12/47.6742/9.0072/Stamen.Terrain&lonlats=8.909225,47.656831;8.912315,47.643186&circlego=9.149895,47.650817,20000&profile=trekking-noferries

grafik

PR #352, opened 2020-12-04
Show inverted polygon for allowed zone

nrenner:
Out of interest I experimented with adding an inverted polygon for the limited area, as a potential addition to #347.

- shows the outside area in a darker shade
- shows a `not-allowed` cursor
- stops clicks on the polygon and thus prevents adding markers outside the area - but cannot avoid dragging markers out

grafik

@bagage what do you think, would that be a useful addition (I don't mind if not)?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -200,6 +200,9 @@ input#trackname:focus:invalid {
 .circlego-draw-enabled {
     cursor: pointer;
 }
+.circlego-outside {
+    cursor: not-allowed;
+}
 
 #map {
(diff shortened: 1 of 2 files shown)


bagage:
Very nice enhancement @nrenner, it isn't much more code and the result looks way better like that!

Do you think we could do a release for that to be available on brouter.de main instance? It would be a really great enhancement for French users :).

nrenner:
> Do you think we could do a release for that to be available on brouter.de main instance? 

Done.

PR #354, opened 2020-12-05
Update dev deps, fix compat issue 

### LINKED ISSUE #353: eslint: location.hostname() is not supported in Opera 71
There is a new compat error related to `location.hostname` in config.js when running `yarn lint`:

>   2:18  error  location.hostname() is not supported in Opera 71  compat/compat
>   2:20  error  location.hostname() is not supported in Opera 71  compat/compat

bagage:
Closes #353 - actually the issue is that `eslint-plugin-compat 3.3.1` changes the behavior for browserlist, and the Opera error is raised. This is not fixed for now.
But since Travis now runs `yarn lint`, we should see it immediately on dep update and see what to do based on that.

Plus I updated some dev dependencies (many code diff is just prettier reformat with new version guidestyle).


bagage:
Updated some deps, but actually I think we should setup renovate-bot instead: even if this is not strictly necessary to update libs, it is better if a bot submits PR (adding links to changelogs etc.), instead of doing it manually once a year or so. WDYT @nrenner?

Cf https://github.com/marketplace/renovate

nrenner:
Thanks for working on this! I neglected updating dependencies and had this on the to-do list for one of the next releases.

Not sure about automated bot updates. They create a lot of noise.

bagage:
I'm using renovatebot on a side [project](https://gitlab.com/bagage/batimap/) and it works great; you can limit noise by:

- limiting to x updates per month or so and
- group updates altogether (say all leaflet related packages update at once for instance, if that makes sense).

What's great about it is that:

- you do not need to check manually for new updates (eg. no need to think of invoking `yarn outdate` and investigating the output).
- links automatically ([example](https://gitlab.com/bagage/batimap/-/merge_requests/763)) to changelog and library homepage, easing a lot the update.
- it can automatically merge it if wished (I personally [let it merge minor/patch versions](https://gitlab.com/bagage/batimap/-/blob/master/renovate.json#L12-15), since the tests in place should be enough to detect a breaking change)

PR #357, opened 2021-01-06
add support for allowed 15 km zone when German localization is active

mjaschen:
Germany is about to restrict movement to a 15 km radius around the home town. This commit enables the "allowed zone" icon for the German localization of BRouter-Web and sets it to a radius of 15,000 meters.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -328,6 +328,10 @@
             circlegoRadius = 20000;
         }
 
+        if (lang.startsWith('de')) {
+            circlegoRadius = 15000;
+        }
+
         if (circlegoRadius != null) {
             circlego = new BR.CircleGoArea(routing, nogos, pois);
             circlego.options.radius = circlegoRadius;


nrenner:
Thanks, I also thought about adopting this for German users.

But to me it is not clear yet, whether the allowed zone is a radius around the home address point or town center point or a rather a buffer around the town border where the 15 km would start outside the town:

> Die Bundesregierung hat „Wohnort“ in ihrem Beschluss nicht näher definiert und damit Spielraum gegeben, ob damit die Adresse oder die Stadt gemeint ist. Auf Nachfrage sagte Kanzlerin Angela Merkel jedoch in der anschließenden Pressekonferenz am Dienstag (05.01.2021), dass damit etwa die Stadt gemeint sei, was den Radius besonders für Menschen in größeren Städten erheblich erweitern würde.

https://www.fr.de/panorama/corona-lockdown-15-kilometer-radius-ausnahmen-deutschland-coronavirus-covid19-bewegungseinschraenkungen-90160053.html

nrenner:
See also [[Talk-de] Karte mit Umkreis von 15km um die Ortsgrenze](https://lists.openstreetmap.org/pipermail/talk-de/2021-January/thread.html#117078).

bueltge:
@mjaschen The simple circle of 15km radius is not the right way, because it is in depend on the Landkreis. Maybe this post is helpful to get a more exactly function - https://www.zeit.de/wissen/gesundheit/2021-01/corona-massnahmen-bewegungsradius-gemeinde-grenze-karte-tool

nrenner:
The Zeit tool uses administrative boundaries, e.g. for [Friedrichshafen](https://www.openstreetmap.org/relation/2783849) (but from the BKG not OSM), which includes the whole countryside and other villages belonging to it.

But I personally would interpret "Wohnort" as only the settled area itself, like 15 km from the city limit sign on. E.g. for [Appenweiler](https://www.openstreetmap.org/way/225617408) I would expect the radius/buffer to be around the residential area of that village. Unfortunately there is no direct definition of such a place area in OSM.

I think we'll have to wait until the first implementation in a state order with a clear definition.

nrenner:
I started collecting information in the wiki page [[de] Corona Bewegungseinschränkung 15 km](https://github.com/nrenner/brouter-web/wiki/%5Bde%5D-Corona-Bewegungseinschr%C3%A4nkung-15-km).

So far it seems each state has its own term and definition.

nrenner:
I made a [summary](https://github.com/nrenner/brouter-web/wiki/%5Bde%5D-Corona-Bewegungseinschr%C3%A4nkung-15-km#zusammenfassung) of what I have found so far.

Number of states implement the 15 km rule as from:
- 4 address
- 4 municipality boundary
- 1 district boundary
- 7 don't implement or leave it to the districts

PR #359, opened 2021-01-15
Determine allowed zone from admin boundaries

### LINKED ISSUE #358: Allowed zone as buffer around admin boundary
Follow-up to #357. See also [[de] Corona Bewegungseinschränkung 15 km](https://github.com/nrenner/brouter-web/wiki/%5Bde%5D-Corona-Bewegungseinschr%C3%A4nkung-15-km).

@mjaschen would you be interested in implementing this? Otherwise I might give it a try today.

I will be looking into potential Overpass API queries.

nrenner:
Closes #358.


mjaschen:
Thanks Norbert, that's great! Just playing around with this feature – call me impressed 👍️

PR #363, opened 2021-01-24
Update docker base image

stefankeidel:
We're running a copy based on Docker and found that some time after 0.13 node 8 stopped working. After some back and forth with different versions, this is one that seems to work. Alpine unfortunately doesn't because git is too outdated for `husky`.


nrenner:
@saesh fine with you?

mjaschen:
(looks good and works 👍️)

nrenner:
Thanks!

PR #366, opened 2021-02-08
Allow to make release without incrementing tag

bagage:
I'm trying to use GitLab for automatic CD - in that manner, I had to change the way I release BRouter on my instance:

- previously I was using `gulp release --minor`
- now, I create a tag via Gitlab UI directly, and a [CI job](https://gitlab.com/bagage/brouter-web/-/blob/dev/.gitlab-ci.yml#L32-36) does the job directly on the given tag (no need to increment it so using `--skipnewtag`). I don't need to commit anything nor to push any tags (since it is now my entrypoint) neither.

Also I moved `bump:html` in `default` task because we want to always override the value in `index.html` with the value from `package.json`. Thoughts?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -210,19 +210,26 @@ var nextVersion;
 var ghToken;
 
 gulp.task('release:init', function (cb) {
+    ghToken = gutil.env.token;
+    if (!ghToken) {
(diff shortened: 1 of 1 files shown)


nrenner:
Looks good to me.

Issue #356, opened 2020-12-20
"voice hint generation" after "Load Track as Route"


EssBee59:
[test_origin_trackonly.gpx.txt](https://github.com/nrenner/brouter-web/files/5721240/test_origin_trackonly.gpx.txt)
[test_origin_osmandStyle.gpx.txt](https://github.com/nrenner/brouter-web/files/5721241/test_origin_osmandStyle.gpx.txt)
[test_afteruploaded_track_osmandStyle.gpx.txt](https://github.com/nrenner/brouter-web/files/5721242/test_afteruploaded_track_osmandStyle.gpx.txt)

Hello,

I am using the Brouter-web to create gpx-tracks to be used in Osmand and set for that the parameter "turnInstructionMode" of the profile to Osmand-style (=3).
By a first generation of a track the voice hints are generated as expected.
But, if I try later to change  the track using the "Load Track as Route" function, the voice hints are sometimes in error (left and right are inverted).
remark:
By the "Load Track as Route" I only upload gpx-tracks (result of export with turnInstructionMode=0), not the gpx track+route as exported with turnInstructionMode=3. (upload of a gpx track+route loads the tour twice!) 
 
I attach some gpx files I created with Brouter-web using the standard "Trekking" profile:
test_origin_trackonly.gpx   ==> turnInstructionMode=0 (using standard "Trekking" profile) 
test_origin_osmandStyle    ==> turnInstructionMode=3 (only this change was made and applied in the "Trekking" profile)
test_afteruploaded_track_osmandStyle ==> exported after "load Track as Route" of the above "text_origin_trackonly" & turnInstructionMode=3
Here the third voice hint (1944 meter after start) is set to "right" instead "left".

The feature "Load Track as Route" is very good, my bikefriends  and myself like and use it: 
Am I doing something wrong?
Or is a modification/work arround possible?
Regards
EssBee

EssBee59:
As remark, each time a wrong turn instruction is generated, 2 identical "trkpt" are found in the track after upload...
example:
Original Track:
   <trkpt lon="8.731992" lat="50.034217"><ele>149.75</ele></trkpt>
   <trkpt lon="8.733453" lat="50.032993"><ele>149.75</ele></trkpt>
   <trkpt lon="8.736330" lat="50.034201"><ele>151.25</ele></trkpt>

Track after upload
   <trkpt lon="8.731992" lat="50.034217"><ele>149.75</ele></trkpt>
[...]

PR #369, opened 2021-02-13
fix for #356, created by @EssBee59

mjaschen:
This pull-request is made be me acting as a proxy for @EssBee59.

### DIFF
--- a/js/plugin/RouteLoaderConverter.js
+++ b/js/plugin/RouteLoaderConverter.js
@@ -305,6 +305,7 @@ BR.routeLoader = function (map, layersControl, routing, pois) {
                     feature = turf.cleanCoords(feature);
                     var lPoints = turf.coordAll(feature);
                     allLinePoints = allLinePoints.concat(lPoints);
+                    console.debug('number of trkpts loaded trackPoints=' + allLinePoints.length);
                 }
             });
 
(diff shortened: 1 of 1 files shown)


EssBee59:
Hello Mjaschen,
Thank you very much for your help!
In the future I will try to learn the procedures to contribute to this product
Regards
Ess Bee

nrenner:
Thanks, will try to have a look at it in the coming week.

js/plugin/RouteLoaderConverter.js
```diff
@@ -328,7 +329,41 @@ BR.routeLoader = function (map, layersControl, routing, pois) {
                 this.getSimplifiedCoords(this._options.simplifyTolerance)
             );
 
+            // use now the middle point between the "simplified" waypoint and the "previous" trkpt in the origin track
+
+            var trackPt_b = turf.cleanCoords(this._trackPoints.geometry);
```
nrenner, line 334: `cleanCoords` is redundant here, it's already done in [getLineStringsFromGeoJSON](https://github.com/nrenner/brouter-web/blob/0ee37f557a9a21a16343a7cf9b5d4d183356b898/js/plugin/RouteLoaderConverter.js#L305)

js/plugin/RouteLoaderConverter.js
```diff
@@ -328,7 +329,41 @@ BR.routeLoader = function (map, layersControl, routing, pois) {
...
+            var trackPt_bis = L.GeoJSON.coordsToLatLngs(trackPt_b.coordinates);
```
nrenner, line 335: Please avoid cryptic variable names with single letters or uncommon abbreviations (what does `b` or `bis` mean?) or just numbering like `buf1`/`2`/`3`, and use [camelCase](https://en.wikipedia.org/wiki/Camel_case) instead of underscore (`_`), e.g. `trackPt_bis` could be `trackLatLngs`

js/plugin/RouteLoaderConverter.js
```diff
@@ -328,7 +329,41 @@ BR.routeLoader = function (map, layersControl, routing, pois) {
...
+            var lookupIndex = 0;
+
             for (var i = 0; i < simplifiedLatLngs.length; i++) {
+                console.debug('simplifiedWP (raw result)=' + simplifiedLatLngs[i] + ': ' + (i - 1));
```
nrenner, line 339: We usually remove `console.debug`/`log` before committing

nrenner (changes requested):
Some minor comments inline.

I guess I would prefer not to include the `filterGPXlines` function:
- The duplicate removal is now obsolete, as `cleanCoords` works as expected in the current Turf version
- If we really wanted to remove route points, I would prefer doing that with the parsed object instead of fiddling with strings, after [DOMParser.parseFromString()](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser#domxrefdomparser.parsefromstring_1):

  https://github.com/nrenner/brouter-web/blob/0ee37f557a9a21a16343a7cf9b5d4d183356b898/js/plugin/RouteLoaderConverter.js#L361-L362

  ```js
  const rteElements = xml.getElementsByTagName('rte');
  for (const rte of rteElements) {
      rte.parentNode.removeChild(rte);
  }
  ```
- But route points should actually be used instead of removed, ideally in a separate PR as not directly related

mjaschen:
If it's ok with @EssBee59 I'd adjust his code an update the PR.

EssBee59:
Hello,

As my changes was "quick and durty", I fully agreed the better code proposed by Norbert!

Glad to see, that "turf.cleanCoords" now works fine!
As explained, the solution I suggested with the middle point of 2 track-points only work properly when the track is not troncated before... (when point of the track are deleted, then the "middle" of 2 trkpt´s is not longer on the route!)

About the use of "route points":
I do not know the use case for this, I am not sure the "turf.simplify" algorithm could help with a route..

I think further...(but NOT SURE...)
"load track as route" is the original name of the function and it is right, but the german translation is not optimal...
"Tour als Route laden" is not a bit confusing...

Regards
Ess bee

nrenner:
> "Tour als Route laden" is a bit confusing...at least with the curent version

Do you mean the word "Tour" is not ideal? Any suggestions?

EssBee59:
It is possibly just a problem with langages (I am french, living in Germany, and some times do not konw longer what is what..)
In the function"Load Track as Route" I would keep for german translation the worf "Track" as I think any biker today know what it is.
"Tour" is I think the general term for a bike voyage (at least in french), it do not describe the type of file that should be uploaded...
And currently only tracks are supported? (possibly route in a future version)
By the way, i prepared a short "user guide" for "Load Track as Route", because my bike friends are not top experts and had a lot of questions aboot the function.
I try to put in attachement, maybe some users could be interested to get an help.
Regards
Ess Bee

nrenner:
>  I would keep for german translation the word "Track" as I think any biker today know what it is.
> "Tour" is I think the general term for a bike voyage (at least in french), it do not describe the type of file that should be uploaded...

Yes, you're right. Feel free to change the [translation](https://github.com/nrenner/brouter-web#translating).

Issue #299, opened 2020-06-06
share + export


ChrizKu:
add share features
- link (optional: custom link)
- <> embed

combine them with existing export features in the menu,
rename it: _share_ instead of _export_ with a suitable icon

link should include parameters #265
alternative to save trips #130

nrenner:
I also thought about having a link with the proposed/edited name in the export dialog, which would help creating a bookmark with the name already set.

> rename it: share instead of export with a suitable icon

Not sure about renaming though, just as "share" is more appropriate for links, it's less so for exporting/downloading/saving the track.

EssBee59:
Is the use of "Custom profile" compatible with this feature?
If not (as I suppose), the feature should be deactivated in that case to avoid errors...

nrenner:
Yes, until #26 is implemented, sharing a link while a custom profile is active should probably at least show a warning.

EssBee59:
Hello,

Sorry, possibly I did not understand wich link exactly should be used for sharing a tour..

I retested today using a custom_profile:
My route:
http://brouter.de/brouter-web/#map=12/49.8632/9.1123/osm-mapnik-german_style&lonlats=8.951797,49.968889;9.138912,49.832226

Now, I unselect the parameter "consider_elevation", the route changes but the URL do not change...
http://brouter.de/brouter-web/#map=12/49.8632/9.1123/osm-mapnik-german_style&lonlats=8.951797,49.968889;9.138912,49.832226 

(in my understanding, only the link to download the gpx changed as it contains the custom_profile/temporary  name)

For that reason, I agree with nrenner not to recommand this method to share routes with friends...

EssBee59:
My error, it seems that the solution proposed was based on the "download link", not the creation link...
Anyway, I think sending the gpx file itself is a more secured way for transfering a tour!

telemaxx:
would be cool if we could give a tour a name, which becomes part of the url. this name should than used for gpx export and saving a bookmark. 
https://www.mtb-news.de/forum/t/brouter-web-fragen-antworten-hilfe-profile-tipps-etc.917910/post-17004934

PR #370, opened 2021-02-14
Export GPX Route via QR Code

mjaschen:
The initial work for this feature was done by @EssBee59. I've integrated the changes and made some adjustments (using a modal dialog instead of opening a new browser window; cleanup code; add support for go-to areas).

- a new button “Generate QR Code” was added to the export dialog
- when that button is clicked a new modal dialog opens
- the new dialog shows a QR Code which contains the “Export as GPX” URL 
- if that URL is opened the GPX file can be downloaded and opened directly on a mobile phone (works on Android and iOS; I've tested it successfully with Firefox and Garmin Connect on iOS 👍️)
- it's possible that the route definition is too long for encoding it into a QR Code, in this case an error message is shown in the modal dialog (happens when a large go-to area is active)
- in the QR Code dialog three buttons are shown – it's possible to switch the QR Code size (helpful if the camera fail to recognize a smaller QR Code)
- the initial size of the QR Code is chosen according the to URL length (< 500 characters => *small,* < 1800 characters => *medium,* else *large*)
- i18n strings are extracted to `en.json`
- the MIT-licensed QR Code library was included (`js/library/qrcode.js`) because it's unavailable as npm package

New Button in Export Dialog:

export-qr-code

QR Code Dialog:

qr-code-dialog

Error when URL is too long:

qrcode-url-too-long

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,3 +130,5 @@ Copyright (c) 2018 Norbert Renner and [contributors](https://github.com/nrenner/
     Copyright (c) 2014-2021 Denis Pushkarev; [MIT License](https://github.com/zloirock/core-js/blob/master/LICENSE)
 -   [regenerator-runtime](https://github.com/facebook/regenerator/tree/master/packages/regenerator-runtime)  
     Copyright (c) 2014-present, Facebook, Inc.; [MIT License](https://github.com/facebook/regenerator/blob/master/packages/regenerator-runtime/LICENSE)
+-   [qrcodejq](https://github.com/davidshimjs/qrcodejs)
+    Copyright (c) 2013-present, @davidshimjs; [MIT License](https://github.com/davidshimjs/qrcodejs/blob/master/LICENSE)
--- a/css/style.css
+++ b/css/style.css
@@ -438,6 +438,22 @@ button.btn {
     margin: -6px 0 6px 20px;
 }
 
+/* QR Code Dialog */
+
+#qrcode-img img {
+    max-width: 100%;
+}
+
+#qrcode-create-large {
+    display: none;
+}
+
+@media screen and (min-width: 500px) {
+    #qrcode-create-large {
+        display: inline-block;
+    }
+}
+
 /*
 * DataTables
 */
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -67,6 +67,7 @@ var paths = {
             'js/Map.js',
             'js/LayersConfig.js',
             'js/router/BRouter.js',
+            'js/library/*.js',
             'js/util/*.js',
             'js/plugin/*.js',
             'js/control/*.js',
--- a/index.html
+++ b/index.html
@@ -539,8 +539,91 @@
                                 >
                                     Export route
                                 </button>
+                                <button
+                                    type="submit"
+                                    class="btn btn-primary"
+                                    data-toggle="modal"
+                                    data-target="#qrcode"
+                                    data-i18n="export.qrcode"
+                                    role="button"
+                                    aria-haspopup="true"
+                                    aria-expanded="false"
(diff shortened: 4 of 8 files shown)


mjaschen:
Want to check out this feature? I've deployed it to my BRouter-Web instance at https://brouter.m11n.de/

EssBee59:
Hello,
the current implementation is vlaid only for brouter-web version smaller 0.17.0  !!!!!
(as the way to dowload the GPX file changed in version 0.17.0 to support "Draw a direct line between two route points")

So please freeze this pull request, I will have a look at the new download and try to adapt this pull request.

@ mjaschen:
Be aware of this problem / current limitation: On your server, the "QRCODE" export do not deliver the right track when a direct line was drawn

Regards

nrenner:
A straightforward way might be to use the client URL (`window.location.href`) for the QR code and add a search/query parameter like `/?export#` to open the export dialog after the route is loaded, e.g. by adding something like this to [index.js](https://github.com/nrenner/brouter-web/blob/5cce9f92c2f8e3d95e0fb23bd33279dc8fb558f5/js/index.js#L313-L314):
```js
        var searchParams = new URLSearchParams(window.location.search);
        if (searchParams.has('export')) {
            routing.once('routing:setWaypointsEnd', function (evt) {
                $('#export').modal('show');
            });
        }
```

mjaschen:
I have been thinking about this feature for the last few days. A user friendly workflow could look like this:

- the QR code contains the current link of the planned route with an additional URL parameter `download=gpx` (`kml`, `xml`, ...)
- if BRouter Web is opened via this link, a modal dialog "Download route or view the planned route" appears with a button for each of the two actions
- a click on "Download" initiates the download, closes the dialog and removes the URL parameter `download=*`.
- a click on "View" closes the dialog and removes the URL parameter `download=*`.

EssBee59:
Norbert,
Thank a lot for your help!!

With your tipp I think it could also be possible to download the GPX file directly on the device that reads the QRCODE (smartphone)?

The other way read the QRCODE and download the GPX file (that I prefer as more confortable) is to use a special app on the android smartphone: also here I think the download can be automated...

@mjaschen:
Do you plan to implement/ adapt the change on the brouter-web side?
If so, I will only change/adapt the app when your part is ready...
Regards
Ess Bee

nrenner:
> With your tipp I think it could also be possible to download the GPX file (=write the track-data) directly on the device that reads the QRCODE (smartphone)?

Don't know what you mean by "directly", but my proposal is that when reading the QR code on the phone, it opens brouter-web in the browser and automatically opens the export dialog. It could also automatically start a specific (e.g. GPX) download after starting brouter-web. The download will be the same as when you open brouter-web on the phone and do a manual export there.

> The other way read the QRCODE and download the GPX file (that I prefer as more confortable) is to use a special app on the android smartphone: also here I think the download can be automated...

As the track formatting now is done in brouter-web JavaScript this could be bit difficult now. You would probably need to make a hybrid app that loads brouter-web in a web view. Or make a minimal headless bundle if Android can run JavaScript.

nrenner:
> the QR code contains the current link of the planned route with an additional URL parameter download=gpx (kml, xml, ...)

For the sharing use case, e.g. embedding the QR code in a blog or website, it would be better to let the reader choose the format.

It might then make sense to have the "generate QR code" button in a separate "share" dialog (see #299) and only open the "export" dialog when reading.

mjaschen:
Yes, that are two valid points 👍 It also keeps redundancies low as the markup for export/download is already available.

Maybe we can add in that case an “Info” section or the like to the export modal which shows one or two sentences about BRouter Web. So people get an idea of where they ended up and what they can do there besides just downloading a GPX file.

PR #372, opened 2021-02-22
Add What's new modal

bagage:
This PR adds a new modal that is displayed to users when they come from an earlier version (that information being stored in localStorage). `What's new?` modal is also always accessible in `About` modal, to see the whole history.

For instance assuming I previously was on 0.13.1, on next launch I will see:

Capture d’écran de 2021-02-22 15-46-56

There are some facts I'm not totally happy with:
- the library I'm using to parse the changelog messes the markdown up quite a bit (links and brackets mainly)
- the changelog is in English only (but I don't think it needs to be translated)

Anyway any feedbacks are welcome as always :).

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -764,3 +764,9 @@ table.dataTable.display tbody tr:hover.selected {
 #selectionText {
     display: none;
 }
+
+.version-new::after {
+    font-family: FontAwesome;
+    content: '\f06a'; /* fa-exclamation */
+    margin-left: 3px;
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,6 +28,8 @@ var rename = require('gulp-rename');
 var browserSync = require('browser-sync');
 var merge = require('merge-stream');
 var babel = require('gulp-babel');
+var marked = require('marked');
+var fs = require('fs');
 
 const server = browserSync.create();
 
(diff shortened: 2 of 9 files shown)


nrenner (changes requested):
Thanks for the PR, I recently thought it would be nice to have some kind of update notification in the client.

I guess we can live with english only and messed up markdown for now (open issue?). Don't know if it would be an alternative to use a proper markdown to HTML renderer and in the client search for the version and split the HTML there ourselves.

But I do hate sites that shove a popup in my face when entering, no matter for what reason. So, could you please change this to something less obstrusive, e.g.
- expand [Message](https://github.com/nrenner/brouter-web/blob/master/js/control/Message.js) class with a `showInfo` type to show a short message and a link to open the modal. Such a message doesn't get in the way and does not require interaction as it disappears with the next route request
- link the version in the Navbar to open the modal and add some "new" badge and highlight

bagage:
I modified the PR according to your feedback @nrenner, thanks!
I dropped the `changelog-parser` dependency in favor of the more popular `marked`, since we don't need much parsing actually it should be OK.

One thing I did *not* do is having a link directly from navbar to Whatsnew modal: when you click on the version, it opens the About modal yet (from which you can go to What's new modal). I wasn't sure about how to do that cleanly and comprehensible by users, so I left it as is. Feel free to modify it if you have an idea for that!

nrenner (approved):

nrenner:
That's fine, it was more like an either/or suggestion.

Thanks!

PR #373, opened 2021-02-24
Add message if no elevation data is available

### LINKED ISSUE #365: communicate lack of elevation data
Latitudes north of 60° seem to have elevation data only between 10° and 30°
east. [0] Which is not you guys’ fault but still a bit frustrating when
planning a trip.

It would be great if the lack of elevation data were indicated somehow. Perhaps
a message could be displayed in the corner of the screen or on the altitude plot.
As it is, it looks rather like a bug in the application itself.

Thanks to @bagage on IRC for nudging me here.

[0] Examples for that boundary:

- bottom 60° N: http://brouter.de/brouter-web/#map=16/59.9985/9.2634/OpenTopoMap,Waymarked_Trails-Cycling&lonlats=9.264672,60.001142;9.266518,59.99742&profile=fastbike
- left 10° E: http://brouter.de/brouter-web/#map=16/60.2772/9.9975/OpenTopoMap,Waymarked_Trails-Cycling&lonlats=9.996827,60.278494;10.007688,60.277739&profile=fastbike
- right 30° E: http://brouter.de/brouter-web/#map=15/69.4232/30.9871/OpenTopoMap,Waymarked_Trails-Cycling&lonlats=31.005484,69.430062;30.96908,69.422591&profile=fastbike

bagage:
Fix #365

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -770,3 +770,16 @@ table.dataTable.display tbody tr:hover.selected {
     content: '\f06a'; /* fa-exclamation */
     margin-left: 3px;
 }
+
+#elevation-chart {
+    position: relative;
+}
+
+#no-elevation-data {
+    z-index: 1000;
+    margin: 0;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+}
--- a/index.html
+++ b/index.html
@@ -1036,7 +1036,11 @@
             </div>
         </div>
 
-        <div class="collapse" id="elevation-chart"></div>
+        <div class="collapse" id="elevation-chart">
+            <span id="no-elevation-data" data-i18n="footer.no-elevation-data" style="display: none"
+                >No elevation data available for this route.</span
+            >
+        </div>
 
         <footer class="flexrow">
             <div id="stats-info" class="flexgrow">
--- a/js/plugin/Heightgraph.js
+++ b/js/plugin/Heightgraph.js
@@ -157,6 +157,16 @@ BR.Heightgraph = function (map, layersControl, routing, pois) {
             }
 
             if (track && track.getLatLngs().length > 0) {
+                // there is no elevation data available above 60°N, except within 10°E-30°E (issue #365)
+                if (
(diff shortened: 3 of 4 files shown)


nrenner:
I would find it nice if there were also a message for partly missing elevation data, e.g. if a route is only half in the no-data-zone, and for cases like in abrensch/brouter#137, i.e. the check would be for one or more undefined `alt`s. What do you think?

bagage:
Yes I initially wanted to handle like this, but did not investigate the difficulty:

image

But maybe for now, a simple label that warns about partial missing elevation data could be enough?

nrenner:
My personal opinion is, that any kind of interpolation is just an assumption and might be completely wrong, so I implemented [gaps](https://github.com/d3/d3-shape/blob/master/README.md#area_defined) on purpose, see also GIScience/Leaflet.Heightgraph#108.

For me, a simple message to inform about partly (or all) missing data, so gaps are not interpreted as mistake, is completely fine.

bagage:
As far as I know, it is BRouter engine that gives use waypoints altitude.

If you do a route from A (no altitude known) to B (altitude known), BRouter will send a list of points without altitude, then some with altitude.
However, if you route from B to A, all points will have an altitude value (the last one known). And there is no property telling you that some elevation data is actually missing/replicated from previous points.

While I could detect the first situation easily, the second one cannot be detected easily (except by patching BRouter to add a property). Any thoughts @nrenner? What would you suggest?

nrenner:
When BRouter interpolates we can't do anything.

I was more thinking about examples where BRouter doesn't, because data is missing at the start of a segment, e.g. your [via on bridge](https://brouter.damsy.net/latest/#map=15/45.1781/4.8188/standard&lonlats=4.814286,45.18863;4.811883,45.184122;4.814329,45.180613) example. My idea was simply to check if **any** `alt` is `undefined` instead of **all**.

But thinking about it, it doesn't look as broken as when there is no data at all, and it shows `Elevation: - m` when hovering as an indication of missing data, so I would wait on feedback for such cases and am merging this now as is.

Issue #286, opened 2020-04-14
Create a logo for Brouter(web)


bagage:
It would be nice to have a proper logo for BRouter to display (for favicon, manifest.json for [progressive web app](https://web.dev/what-are-pwas/), maybe even the [Android application](https://play.google.com/store/apps/details?id=btools.routingapp&hl=en)).

WDYT @nrenner, @abrensch?

I'm not able to create a proper logo personally*, but I believe we could find someone which could help for that.

* my own made logo is :fearful::
![icon-192](https://user-images.githubusercontent.com/1451988/79249469-9855e800-7e7d-11ea-833b-62aadcdcb7a8.png)

nrenner:
Yes, having a logo would be nice, I personally have this in mind for a 1.0 version once more pressing issues are done.

One idea for a temporary solution I have, would be to use the current BRouter app image and maybe add some "-Web" to it, which often is some kind of globe icon? But I don't know if having a common base brand is a good idea or desired.
![BRouter logo](https://lh3.googleusercontent.com/fCLMs01-CotQCyqyjnsUyHxcLvWgFyxo_izZlurV1Hn5wEMWcd3lCCf-JGuVUi5Klzw=s180)

A bike symbol might not fully represent what BRouter stands for, as abrensch seems more interested in electric car support lately and we offer all kinds of profiles.

bueltge:
I'm not a designer, but maybe the follow examples are useful.

![image](https://user-images.githubusercontent.com/133425/79978098-a15b4080-849f-11ea-96f3-96a43bc488c2.png)

nrenner:
Thanks for the suggestion!

nrenner:
@bueltge I like the core idea of using a curved arrow as a symbol for a route.

However, the proposed icon very much looks like a [random](https://fontawesome.com/icons/random?style=solid)/shuffle symbol, e.g. as used in music players like Spotify for random play, which is a bad analogy for routing:

![random (Font Awesome)](https://user-images.githubusercontent.com/1092030/82735946-fcde4f80-9d25-11ea-9573-176fae68fcd7.png)

I would make one arrow shaded or gray, it would better reflect that a router chooses one optimal route, and less ideal alternatives, e.g. like the [duotone](https://fontawesome.com/icons/random?style=duotone) variant:

![random duotone (Font Awesome)](https://user-images.githubusercontent.com/1092030/82735966-4333ae80-9d26-11ea-8c7e-c2525922bf4c.png)

[...]

PR #382, opened 2021-03-08
WIP: Add PWA/TWA required files

bagage:
This aims to make BRouter a [Progressive Web App](https://web.dev/progressive-web-apps/) (PWA) and [Trusted Web App](https://developers.google.com/web/android/trusted-web-activity/quick-start) (TWA).

There are various things to finalize:

- mobile version could be improved (#296), not mandatory
- have a proper logo (#286). For now and test purposes, I'm using my placeholder. The maskable icon was made with https://web.dev/maskable-icon/
- define BRouter [theme](https://developer.mozilla.org/en-US/docs/Web/Manifest/theme_color) and [background](https://developer.mozilla.org/en-US/docs/Web/Manifest/background_color) colors. For now I've set both to `#343a40`.
- I deployed `manifest.json` and `.well-known/assetlinks.json` to my instance. You can try TWA from [this APK](https://gitlab.com/bagage/brouter-web/uploads/a03409ed428fbfd112931a89be7d90cf/BRouter.apk) (generated via https://www.pwabuilder.com/publish)
- the best (I think) solution would be to integrate the TWA directly into the existing BRouter application. But a separate app can be used too (you only have to update `assetlinks.json` accordingly)
- to generate the APK, you can use pwabuilder.com but I believe it's better to use `bubblewrap` directly and commit required files in brouter or brouter-web repository
- some files might be removed from the repository, or not, I'm unsure yet.


nrenner:
> This aims to make BRouter a Progressive Web App (PWA) and Trusted Web App (TWA).

When you write "BRouter" in the description and also in the manifest, that should probably be "BRouter-Web"?

I have a rough idea what a PWA is, but I'm not familiar yet with TWA. So a PWA can install on the home screen but not go full screen, for that you need a TWA installed as an APK?

> You can try TWA from this APK 

That link doesn't work.

> mobile version could be improved (#298), not mandatory

You proabably mean #296?

> the best (I think) solution would be to integrate the TWA directly into the existing BRouter application.

What files would need to be integrated there, would that be a one-time thing or for every BRouter-Web release? Do we need a separate icon when using a single APK?

Do all those files need to be in the root folder? We already have too many files in the project root and I would prefer to put anything that doesn't need to be there into some subfolder, e.g. the images into a `resources` subfolder.

bagage:
> When you write "BRouter" in the description and also in the manifest, that should probably be "BRouter-Web"?

Yes indeed. Will rename it accordingly.
> 
> I have a rough idea what a PWA is, but I'm not familiar yet with TWA. So a PWA can install on the home screen but not go full screen, for that you need a TWA installed as an APK?

I think there are 2 main differences:
- a TWA can be published on the play store or alternative apps markets, so you can have a "native experience" with the app.
- PWA/TWA are webviews, so it's basically Chrome or Firefox which run underneath. If you use a TWA, the browser navigation bar is hidden, allowing to gain some height space:

diff

> 
> > You can try TWA from this APK
> 
> That link doesn't work.

Indeed, my bad. [Here it is](https://gitlab.com/bagage/brouter-web/uploads/a03409ed428fbfd112931a89be7d90cf/BRouter.apk).
> 
> > mobile version could be improved (#298), not mandatory
> 
> You proabably mean #296?


Yes indeed.

> 
> > the best (I think) solution would be to integrate the TWA directly into the existing BRouter application.
> 
> What files would need to be integrated there, would that be a one-time thing or for every BRouter-Web release? Do we need a separate icon when using a single APK?

The APK can be (re)generated from the `manifest.json` (which itself links to icons that must be accessible online). Rebuilding the APK is only required when the manifest and/or icons change. Otherwise the APK is basically a webview pointing to the website, so whenever a new BRouter-web release is deployed, the APK is just up-to-date as well.

> 
> Do all those files need to be in the root folder? We already have too many files in the project root and I would prefer to put anything that doesn't need to be there into some subfolder, e.g. the images into a `resources` subfolder.

No they can be put elsewhere. Some have to be embedded in Brouter releases (`pwabuilder-sw.js`, maybe `favicon.ico`), the others can be put in a separate `resources` subfolder.

mjaschen:
There are some hardcoded things (namespace/fingerprint) in `assetlinks.json`. 

What does that mean for BRouter-Web instances which are deployed elsewhere?

bagage:
@mjaschen I'm not 100% sure about all PWA related stuff, but from what I understand:

`assetlinks.json` is meant to be deployed on websites ([here](https://brouter.damsy.net/.well-known/assetlinks.json) for my instance). It lists apps that are trusted for this website, eg. any APK with identifier `net.damsy.brouter` and signed with `9F:...:D1` (which is the APK I generated above) hash signature is a trusted app (the only consequence I know is that the browser nagivation bar is hidden). 
But if you do not deploy `assetlinks.json` on your website and/or if you remove that entry from the file, then my APK won't be trusted on your instance and thus it will simply be a PWA. You can list many apps in that file - but this file could also be out of brouter repository, I don't know best practice about that yet.

nrenner:
>  If you use a TWA, the browser navigation bar is hidden, allowing to gain some height space

Doesn't the `"display": "standalone"` or `fullscreen` in the PWA manifest already do that (without TWA)?
https://web.dev/add-manifest/#display

bagage:
From my tests, at least with Firefox, it does not hide the navigation bar without TWA, but I might be missing something.

Phyks:
> From my tests, at least with Firefox, it does not hide the navigation bar without TWA, but I might be missing something.

As far as I know, Firefox support for PWA can be quite limited on Android devices. Sadly, Chrome should be better for tests here.

bueltge (approved):
Would be really helpful for the mobile usage.

bueltge:
@bagage Great!
Would you add an option to change the static strings inside the manifest.json dynamically for deployment?
Hint https://stackoverflow.com/questions/50086761/dynamic-icon-pwa-manifest

bagage:
Closing this as I don't plan to finish it anytime soon.

PR #383, opened 2021-03-10
Feat/improve mobile

### LINKED ISSUE #296: improve mobile responsivity

bagage:
Not totally finished.

Closes #296

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -242,10 +242,6 @@ input#trackname:focus:invalid {
     display: flex;
 }
 
-#profile_buttons {
-    padding-top: 4px;
-}
-
 /* track messages (data tab) */
 #tab_data,
 .CodeMirror {
--- a/css/style.css
+++ b/css/style.css
@@ -788,3 +788,136 @@ table.dataTable.display tbody tr:hover.selected {
 .leaflet-control-layers-selector[disabled] ~ span {
     color: #777;
 }
+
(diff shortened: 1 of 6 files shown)


bagage:
Should be good to be reviewed @nrenner. Only one item left, but I opened a upstream issue so can be done separately.

nrenner:
Thanks!

Issue #106, opened 2018-03-16
POI layer


nrenner:
Requests to show Points Of Interest (POI) on the map for tour planning:
- youth hostels ([Hostel overlay (Groups)](https://groups.google.com/d/msg/osm-android-bikerouting/OyeMk1TzofM/nopPpA83czgJ))
- Historic objects (#105)

It would be nice to have a general Overpass API POI layer with some UI to select/filter for specific POIs/categories. Not aware of any plugin for the UI part though.

Maybe utilizing tag hierarchies, icons and translations for selection and popup content from POI apps or [editor presets](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md)?

## Overpass layers

- [leaflet-nectarivore](https://github.com/osmlab/leaflet-nectarivore)
- [overpass-layer](https://github.com/plepe/overpass-layer) + [overpass-frontend](https://github.com/plepe/overpass-frontend)
- [leaflet-overpass-layer](https://github.com/GuillaumeAmat/leaflet-overpass-layer)

## POI apps

- [OpenStreetBrowser](https://github.com/plepe/OpenStreetBrowser)
- [osm24](https://github.com/dotevo/osm24/) - http://osmapa.pl/osm24
- [Unterkunftskarte](http://unterkunftskarte.de/#14/51.0700/13.7299) - no repository
- [openpoimap](https://github.com/marczoutendijk/openpoimap) - http://openpoimap.org/, OpenLayers

## Router with POIs examples

- [Openrouteservice Maps](https://maps.openrouteservice.org/) - pin button on the right, own DB
- [Komoot](https://www.komoot.de/plan/) - click in search bar, closed source
- [Naviki](https://www.naviki.org/de/naviki/route-planen/) - button top right, closed source

hugobqd:
When I prepare a mountain bike ride with brouter-web, I always look for drinking water on my way with overpass turbo. It would be awesome to have it and other requests into brouter-web

This are my four request : 
`[man_made=water_tap]
  [amenity=drinking_water]
  [amenity=toilets][drinking_water!=no]
  [drinking_water=yes]`

stesie:
Hej, just wanted to let you know that I've started working on this.

So far you can enable an extra overpass layer + either enter a custom overpass query (updated on textarea blur) or choose from the (yet little) template list. The marker popup definitely needs more content, ... and generally early feedback is welcome :-)

[...]

Issue #306, opened 2020-06-08
Reusing translations from other projects?


nrenner:
PR #304 introduces dynamic translations of highway, surface and smoothness tag values. Some with underscore were added manually to the `en.json` and got removed by `i18next`:

https://github.com/nrenner/brouter-web/blob/f37489cf265ab5e4b24224ac135ed1d66cc351eb/locales/en.json#L196-L207

They would need to be added to [keys.js](https://github.com/nrenner/brouter-web/blob/master/locales/keys.js), but I left them off for now, as I think we should not translate them ourselves, but ideally reuse existing translations, some OSM editor must already have those.

On Transifex we are in the OpenStreetMap organisation, and e.g. the Vespucci editor is also part of that.

Can translations be shared between projects on Transifex (haven't done any research yet)?

nrenner:
The concepts for reusing translations on Transifex I found are [Sharing Translation Memory](https://docs.transifex.com/translation-memory/sharing-tm) and [Glossary](https://docs.transifex.com/glossary/glossary). These are presented as suggestions, so still invole translators. Translation Memory also allows to automatically translate matching texts, but I'm a bit sceptical about that.

I haven't found anything that would allow to reference or link to specific translation keys from other projects.

i18next has the concept of [Namespaces](https://www.i18next.com/principles/namespaces) to load from separate files.

So my idea would be extract the required keys directly from the translation files of another project into separate files and not go through Transifex.

Some candidates are:
- [Beautified JOSM preset](https://github.com/simonpoole/beautified-JOSM-preset): [PO format](https://github.com/simonpoole/beautified-JOSM-preset/tree/master/i18n), (preset: [master_preset.xml](https://github.com/simonpoole/beautified-JOSM-preset/blob/master/master_preset.xml)), [smoothness only as descriptions](https://github.com/simonpoole/beautified-JOSM-preset/blob/d77ccd9475e07645c15e8bc1cb09d71289273dd7/master_preset.xml#L35-L46) ¹
[...]

PR #384, opened 2021-03-19
Add (Overpass based) POI layer

stesie:
This adds a (yet pretty simple) Overpass API based POI layer.  To achieve this it integrates the MIT-licensed overpass-layer module, which runs overpass queries for the full bounding box (i.e. no "search along the route"), and uses a map marker quite like the existing POI marker (but with `fa-search` icon instead of `fa-star`). The popups list the OSM-style key/value pairs without much interpretation.

Overpass layers can be added in two ways:
* as default layers, with .geojson files residing statically below layers/overpass/
* as custom layers, that can be added dynamically by the user

Regarding the default layers I was wondering how many of those layers should be added (or if any at all), and how specific they should be e.g. to cycling needs (e.g. water sources of different kinds) or general purpose (one special kind of amenity per layer). This PR is currently leaning towards the latter and just adds a whole lot of amenities, shops & tourism osm features --- likely too many.

With the custom layers dialog I was also wondering, how things can be achieved in a user friendly way ... so far it's pretty cryptic (mainly since I refrained from just putting documentation into it or adding actual labels to the inputs).  So instead of an URL you can put a Overpass Query (like e.g. `(node[shop]['diet:vegan']['diet:vegan'!=no]; way[shop]['diet:vegan']['diet:vegan'!=no];);`) in the URL input field, provide a layer name and click "add overpass query".

Both kinds of layers then can be checked on and off in the layer control like normal.

For the moment I'd consider this a minimal baseline, on which sub-sequent features could be built -- like choosing a color for the layer (which could be of interest to the gpx import layers as well)

refs #106

### DIFF
--- a/js/LayersConfig.js
+++ b/js/LayersConfig.js
@@ -1,4 +1,7 @@
 BR.LayersConfig = L.Class.extend({
+    overpassFrontend: new OverpassFrontend(
+        (BR.conf.overpassBaseUrl || '//overpass-api.de/api/interpreter').replace('?data=', '')
+    ),
     defaultBaseLayers: BR.confLayers.defaultBaseLayers,
     defaultOverlays: BR.confLayers.defaultOverlays,
     legacyNameToIdMap: BR.confLayers.legacyNameToIdMap,
@@ -251,6 +254,19 @@ BR.LayersConfig = L.Class.extend({
             if (props.subdomains) {
                 layer.subdomains = props.subdomains;
             }
+        } else if (props.dataSource === 'OverpassAPI') {
+            layer = new OverpassLayer({
+                overpassFrontend: this.overpassFrontend,
+                query: props.query,
+                minZoom: undefined,
+                feature: {
+                    title: '{{ tags.name }}',
+                    body:
+                        '<table class="overpass-tags">{% for k, v in tags %}{% if k[:5] != "addr:" %}<tr><th>{{ k }}</th><td>{{ v }}</td></tr>{% endif %}{% endfor %}</table>',
+                    markerSymbol:
+                        '<img anchorX="13" anchorY="42" width="25" height="42" signAnchorX="0" signAnchorY="-30" src="dist/images/marker-icon.png">',
+                },
+            });
         } else {
             // JOSM
             var josmUrl = url;
--- a/js/LayersConfig.js
+++ b/js/LayersConfig.js
@@ -10,6 +10,8 @@ BR.LayersConfig = L.Class.extend({
 
     initialize: function (map) {
         this._map = map;
+        this._overpassLoadingIndicator = new BR.Message('overpass_loading_indicator', { alert: false });
+        this._overpassActiveRequestCount = 0;
 
         this._addLeafletProvidersLayers();
         this._customizeLayers();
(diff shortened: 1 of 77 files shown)


stesie:
Just willing to give it a try?  I've deployed this to my personal BRouter-Web instance at https://brouter.brokenpipe.de/
(however it has segment data for around Germany only)

js/LayersConfig.js
```diff
@@ -169,6 +172,21 @@ BR.LayersConfig = L.Class.extend({
         return result;
     },
 
+    createOverpassLayer: function (query) {
+        return new OverpassLayer({
+            overpassFrontend: this.overpassFrontend,
+            query: query,
+            minZoom: undefined,
```
nrenner, line 179: We really should set a minZoom and not allow users to query the whole world and put unnecessary load on the Overpass API. My suggestion would be `12`.

nrenner (changes requested):
Thanks for the PR!

Some additional thoughts for potential (later) improvements (not critical for merging):
- The queries only include nodes and ways, which will miss some objects mapped as relations, e.g. this `camp_site` mapped both as [site](https://www.openstreetmap.org/relation/4741639) and [multipolygon](https://www.openstreetmap.org/relation/4741633) relation. Don't know about performance implications of always using the [`nwr` shorthand](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#The_Query_Statement). Otherwise [taginfo](https://taginfo.openstreetmap.org/tags/?key=tourism&value=camp_site#overview) shows relation usage per tag.
- I personally find the default styling of nodes as both CircleMarker and icon Marker superfluous and a bit too much
- My impression is, that the default Overpass instance is slow and overloaded these days, perhaps we should use an [alternative instance](https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances)?
- Some kind of loading indicator would be nice

nrenner:
> Regarding the default layers I was wondering how many of those layers should be added 

Choosing relevant layers is hard. From a maintenance perspective I guess more is better, so I don't get frequent requests to add yet another layer.

> With the custom layers dialog I was also wondering, how things can be achieved in a user friendly way

Not sure. Maybe by putting the current input placeholder as help text below like in the profile options tab and adding an example for Overpass, that can be copied this way (unlike the placeholder). Don't know if something like [overpass-wizard](https://github.com/tyrasd/overpass-wizard) would help.

stesie:
I completely agree with the minZoom requirement. `overpass-layer`'s default actually is 16, which felt a bit too strict. Yet no limit at all might well be too lax :)

Regarding `camp_site` (and others) being mapped as relations, when creating all the geojson files I had a close look to OSM wiki, e.g. https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dcamp_site, which states that `tourism=camp_site` should not be used on ways and relations.  And so on for the other files.  But as you pointed out, there are e.g. camp sites that are mapped as relations, and they probably should be found as well. There's even a bunch of `amenity=bench` which are mapped as relations. Hence I've just changed all the search defintions to always use `nwr[]` no matter what. (unaware of the performance impact, but it doesn't feel worse with a couple of tries)

Regarding the marker symbol I also agree, so plain nodes are now depitected by a simple circle marker, and both areas as well as multi polygon relations by their outline.

I like the idea to split the custom layers dialog's placeholder into label and helptext parts, especially to keep the helptext visible. So I've just quickly changed that as well.

Regarding the loading indicator I need to have a closer look someday. To be honest I've been missing it myself from time to time already. Yet from a brief look into the documentation it doesn't (yet) emit the signals necessary to achieve this.

nrenner:
> Regarding the marker symbol I also agree, so plain nodes are now depitected by a simple circle marker, and both areas as well as multi polygon relations by their outline.

Not sure, but I would have tended to do it the other way round and just not show the CircleMarker for nodes, but keep both marker and outline for ways/areas:
- especially buildings get reduced to a few pixels and are harder to see when zooming out, so a marker helps, but the outline is useful additional information as well (Overpass Turbo has an option to turn small features into CircleMarkers on lower zooms)
- a future improvement might be to distinguish multiple Overpass layers by specific icons in the markers

Don't know how to do that with overpass-layer though, there seems to be a `nodeFeature` style option, see [parks demo](https://github.com/plepe/overpass-layer/blob/6416b79390c68d336c26f0327cb13b02b42b1839/demo/parks.js#L9) and [overpass-frontend OverpassNode.js](https://github.com/plepe/overpass-frontend/blob/7bb8b002e8d7ec4f6f7bbdf8240218ebc7f2b6b0/src/OverpassNode.js#L115-L123).

> Regarding the loading indicator [...] Yet from a brief look into the documentation it doesn't (yet) emit the signals necessary to achieve this.

In the [parks demo](https://github.com/plepe/overpass-layer/blob/6416b79390c68d336c26f0327cb13b02b42b1839/demo/parks.js#L23-L28) I just saw `onLoadStart` and `onLoadEnd` handlers.

Let me know if you want to leave it as is for now or add some more changes to this PR. I'm fine with merging as is.

nrenner (approved):

stesie:
I've just added a simple loading indicator, based on the existing messaging approach.  Yet I didn't want to accidentally wipe out possibly visible error messages, hence I've wrapped the `#message` container to allow for a secondary alert to be shown right below it.

And thanks for pointing out those two event handlers exist, I wonder where I was looking at yesterday without noticing them. Whatever ...

Regarding multiple Overpass layers, so far I have been thinking of using separate colours instead of icons, so users could control the colour of either markers, circles or polygons. Having a colour picker feels more intuitive then adding some kind of icon selector, and also would be reusable for e.g. imported gpx route layers.

nrenner:
Thanks for your contribution!

nrenner:
I made some minor additions in #389.

> Regarding multiple Overpass layers [...]. Having a colour picker feels more intuitive then adding some kind of icon selector, 

My idea was not to let the user select, but to extend the overpass layer definitions with icons that would show both in the optional layers tree and in the markers for a selected layer. Ideally some existing icon associations and also translations could be used, e.g. from editor presets, see also #306.

Example from [id-tagging-schema presets.json](https://github.com/openstreetmap/id-tagging-schema/blob/9fdecc983443ce75185c2923caae9f7703f2cae4/dist/presets.json#L22359-L22360):
```json
    "tourism/camp_site": {
        "icon": "maki-campsite",
```
which would be [maki campsite-15.svg](https://github.com/mapbox/maki/blob/master/icons/campsite-15.svg): maki-campsite

Issue #385, opened 2021-03-22
Auto-focus primary action button in dialogs


mjaschen:
When using the keyboard shortcut to delete a route the confirmation can't be submitted by pressing `Enter`.

`Esc` for dismissing the dialog without doing any changes works on the other hand.

I'd like to be able to confirm the primary action (delete the route) by pressing `Enter` after invoking the modal dialog via `Backspace` key.

*bootbox* has no options to control this behavior. Starting with version 5.4.0 bootbox allows running a callback when a dialog is shown. This allows the set the focus to the primary dialog button. 

I've updated the bootbox requirement from `5.3.4` → `5.5.2` (current version) and added the callback locally (works fine).

Is that a feature more people than me are interested in? If so, I’d format a pull request.

nrenner:
I guess it would make sense for this one specific dialog, but not for others, i.e. the add POI dialog.

rkflx:
> Is that a feature more people than me are interested in?

+1

I've been running a (probably) similar patch (bootbox upgrade and handling of Enter key, plus some changes for dialogs not covered by the bootbox change) since a couple of months. I would not want to use BRouter-Web without it anymore.

Did not notice any negative effects on the POI dialog so far.

mjaschen:
> this one specific dialog, but not for others

I implemented it just for this one dialog actually ;-)

I'll submit a PR later.

PR #387, opened 2021-03-23
auto-focus primary button in "delete route" dialog

mjaschen:
After opening the "delete route" dialog the primary action button gets the
current focus. So it's possible to confirm resetting the route by pressing
"Enter".

bootbox had to be updated to to achieve this (in prior versions of bootbox the
needed callback didn't exist).

see #385

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -153,6 +153,9 @@
                     },
                 ],
                 value: ['route'],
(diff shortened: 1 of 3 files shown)


nrenner:
Thanks!

PR #388, opened 2021-03-23
Polishes whatsnew modal

### LINKED ISSUE #386: What's new follow-up
Issues after #372:

- the [switch statement](https://github.com/nrenner/brouter-web/blob/d3e344fd25b8535b68dcc9480fd9c9871dc78d40/js/control/Message.js#L17-L26) in the Message class is missing a `break;` for each case, so errors and warnings now show as info
- new users without `changelogVersion` set, or in incognito mode, shouldn't get a "what's new" message, but do (full changelog)

bagage:
Closes #386.

### DIFF
--- a/index.html
+++ b/index.html
@@ -910,6 +910,7 @@
                 <div id="map" class="leaflet-sidebar-map">
                     <div id="notification_jar">
                         <div id="message"></div>
+                        <div id="whats_new_message"></div>
                         <div id="overpass_loading_indicator"></div>
                     </div>
                     <div id="preview" hidden data-i18n="map.preview">Preview</div>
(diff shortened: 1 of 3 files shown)


nrenner:
I added `localStorageAvailable` checks in e9540170a79c3b1d5ea81db4243db4fd60cbf89c.

But when testing I get the full changelog shown after a blink, maybe because opening the modal closes the message and triggers `onClosed`?

E.g. CHANGELOG.md with added:
```
## 0.16.0 (2021-03-25)

-   Test

```
and in the browser console:
```js
localStorage.setItem('changelogVersion', "0151-2021-01-25")
```

bagage:
I'm not sure what is happening. This is what happens for me:

Capture d’écran vidéo de 26-03-2021 10:29:34_cropped

nrenner:
It's also obvious from your screencast that the message disappears when the modal opens. I tried to find out where the closing event is coming from by setting a breakpoint in the `onClosed` handler, but the origin was just some transitionend event that didn't help much.

A workaround might be to do the `onClosed` stuff when closing to the modal instead of the message?

PR #390, opened 2021-03-24
Route Loader Keyboard Shortcut

mjaschen:
This pull request enables opening the “import as route” by pressing `r` (for “route”) on the keyboard.

Additionally, an unused option value is removed from the `cleanup()` method.

### DIFF
--- a/index.html
+++ b/index.html
@@ -78,12 +78,14 @@
                             aria-haspopup="true"
                             aria-expanded="false"
                             data-i18n="[title]navbar.load.tooltip"
-                            data-i18n-options="{
-                                'tracksAction': '$t(navbar.load.tracks)',
-                                'tracksKey': 'O',
-                                'nogosAction': '$t(navbar.load.nogos)',
-                                'nogosKey': '$t(keyboard.shift)+O'
-                            }"
+                            data-i18n-options='{
+                                "tracksAction": "$t(navbar.load.tracks)",
+                                "tracksKey": "O",
+                                "trackAsRouteAction": "$t(trackasroute.title)",
+                                "trackAsRouteKey": "$t(keyboard.shift)+O",
+                                "nogosAction": "$t(navbar.load.nogos)",
+                                "nogosKey": "$t(keyboard.shift)+N"
+                            }'
                             title="Load route"
                         >
                             <span class="fa fa-lg fa-cloud-upload" aria-hidden="true"> </span>
--- a/js/plugin/NogoAreas.js
+++ b/js/plugin/NogoAreas.js
@@ -5,6 +5,7 @@ BR.NogoAreas = L.Control.extend({
                 enable: 78, // char code for 'n'
                 disable: 27, // char code for 'ESC'
             },
+            import: 78, // char code for 'n'; used in conjunction with 'shift'
         },
     },
 
@@ -147,6 +148,13 @@ BR.NogoAreas = L.Control.extend({
         if (!BR.Util.keyboardShortcutsAllowed(e)) {
             return;
         }
+
+        if (true === e.shiftKey && e.keyCode === this.options.shortcut.import) {
+            $('#loadNogos').modal('show');
+
+            return;
+        }
+
         if (e.keyCode === this.options.shortcut.draw.disable && this.button.state() === BR.NogoAreas.STATE_CANCEL) {
             this.stopDrawing(this.button);
         } else if (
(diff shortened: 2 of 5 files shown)


rkflx:
Did you test for shortcut conflicts?

Any tooltip changes needed?

rkflx:
> Any tooltip changes needed?

BTW, tooltips in the top bar might be broken currently, possibly due to 5b52abc1.

bagage:
Good catch @rkflx, I pushed a fix https://github.com/nrenner/brouter-web/commit/17bbef17bdc0f824e9d316fe7f36896ac4823e65 for that, thanks!

mjaschen:
> Did you test for shortcut conflicts?

Yes, but after a second check I noticed, that `r` is indeed already used for `reverse route` 🤦‍♂️  I didn't notice that, because importing a track as route is usually done without an existing route (which means that pressing `r` has no effect in this case). Thanks for pointing this out.

My next proposal would have been `Shift-O` but this shortcut is already used (“import no-go area”)

The list of all keyboard shortcuts which are currently defined (maybe also a good topic for the upcoming documentation efforts):

|Shortcut|Description|
|:-:|---|
| `D` | enter draw mode |
| `Esc` | exit draw mode or close dialog |
| `Z` | remove last point from route (“undo”) |
| `M` | hide route temporarily (as long as key is down) |
| `P` | insert „Point of Interest“ (`Esc` exists this mode) |
| `+`, `-` | zoom map |
| `←`, `→`, `↓`, `↑` | move map |
| `L` | center current position (geolocation browser-feature) |
| `F` | search for locations |
| `R` | reverse route |
| `N` | draw „No Go Area“ |
| `Backspace` | delete route |
| `C` | toggle through color-coding modes |
| `S` | activate/deactivate Strava layer (if configured) |
| `E` | show elevation profile |
| `G` | switch routing profile |
| `X` | export route |
| `O` | load track |
| `Shift-O` | load “No Go Area” |
| `H` | show “About” dialog |
| `T` | toggle sidebar |
| `Shift-T` | switch tabs in sidebar |
| `I` | draw allowed zone |

New ideas for the “import as route” shortcut: `Shift-R` or `U`. Maybe there are other suggestions?

rkflx:
> New ideas for the “import as route” shortcut: `Shift-R` or `U`.

How about `N` and `Shift+N` for draw/load „No Go Area“, so we could have `O` and `Shift+O` for „Import as Track/Route“ respectively?

---

> Good catch @rkflx, I pushed a fix 17bbef1 for that, thanks!

Thanks @bagage!

mjaschen:
> How about N and Shift+N for draw/load „No Go Area“, so we could have O and Shift+O for „Import as Track/Route“ respectively?

+1

I'd like that. Maybe apart from the fact, that we've to change an established shortcut. But I personally wouldn't mind to do such a change 🤷🏻 

@nrenner do you have an opinion on this?

nrenner:
It would be a good improvement, so I guess the change is justified.

mjaschen:
Great! I'll update the pull request.

mjaschen:
Shortcuts/tooltips were updated. 

Updated keyboard shortcuts overview:

|Shortcut|Description|
|:-:|---|
| `D` | enter draw mode |
| `Esc` | exit draw mode or close dialog |
| `Z` | remove last point from route (“undo”) |
| `M` | hide route temporarily (as long as key is down) |
| `P` | insert „Point of Interest“ (`Esc` exists this mode) |
| `+`, `-` | zoom map |
| `←`, `→`, `↓`, `↑` | move map |
| `L` | center current position (geolocation browser-feature) |
| `F` | search for locations |
| `R` | reverse route |
| `N` | draw „No Go Area“ (`Esc` exists this mode) |
| `Shift-N` | load “No Go Area” |
| `Backspace` | delete route |
| `C` | toggle through color-coding modes |
| `S` | activate/deactivate Strava layer (if configured) |
| `E` | show elevation profile |
| `G` | switch routing profile |
| `X` | export route |
| `O` | load track |
| `Shift-O` | load track as route |
| `H` | show “About” dialog |
| `T` | toggle sidebar |
| `Shift-T` | switch tabs in sidebar |
| `I` | draw allowed zone |

nrenner:
Thanks!

nrenner:
I added the overview to the wiki:
https://github.com/nrenner/brouter-web/wiki/Keyboard-shortcuts

PR #395, opened 2021-04-01
Improve keyboard handling and dialog UIs

rkflx:
Let's upstream some local changes and fix a couple of regressions in the process, mostly related to keyboard shortcuts and various dialogs:

- Add more exceptions to .prettierignore
- Fix broken Return key in export route dialog
- Fix loading no-go areas
- Show missing filename in load no-go areas dialog
- Allow Return key to accept dialog when loading no-go areas or track as route
- Set focus in modals to first input field by default
- Select parts of trackname in export dialog for easier overwriting

(See commit messages for more details.)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,11 +1,14 @@
 LICENSE
+Dockerfile
 dist/
 *.zip
 yarn.lock
 .idea
 .gitignore
+.eslintignore
 .prettierignore
 .tx/
 layers/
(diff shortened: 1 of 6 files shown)


nrenner (commented):
> 5b52abc broke importing no-go areas, in particular after choosing a file
and clicking on "Load" nothing would happen.

Oops.

Thanks!

PR #397, opened 2021-04-03
Use L.Marker in Overpass layer + maki/temaki icons

stesie:
As suggested in https://github.com/nrenner/brouter-web/issues/384#issuecomment-805719150, this PR

* uses `L.Marker` instead of `L.Circle` to mark nodes found via Overpass
* adds icon descriptors from [id-tagging-schema presets.json](https://github.com/openstreetmap/id-tagging-schema/blob/9fdecc983443ce75185c2923caae9f7703f2cae4/dist/presets.json#L22359-L22360) to layers config
* shows maki/temaki icons, according to those
* falls back to fa-search icon (custom layers always get the fallback icon currently)

#### caveats

1. overpass-layer currently has an issue with using "nodeFeature=Marker". Initially it works fine - yet after zooming out it reprocesses its' map features and wants to call `setStyle` on it (to possibly update styling info). However `L.Marker` doesn't provide that function (as opposed to `L.Circle` or `L.CircleMarker`). I've reported this upstream at https://github.com/plepe/overpass-layer/issues/7
For the moment I added some kind of a work-around: if it's just a node, the circle color is changed to alpha=0, i.e. made transparent.

2. the id-tagging-schema references some icons from font-awesome 5. Yet we're using version 4.7 currently, which is missing most of the icons. For the moment those `fas-xxx` icons are *not* handled, i.e. the fallback icon (fa-search) is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,3 +132,7 @@ Copyright (c) 2018 Norbert Renner and [contributors](https://github.com/nrenner/
     Copyright (c) 2014-present, Facebook, Inc.; [MIT License](https://github.com/facebook/regenerator/blob/master/packages/regenerator-runtime/LICENSE)
 -   [overpass-layer](https://github.com/plepe/overpass-layer)
     Copyright (c) 2020 Stephan Bösch-Plepelits; [MIT License](https://github.com/plepe/overpass-layer/blob/master/LICENSE)
+-   [maki](https://github.com/mapbox/maki)
+    [CC0-1.0 License](https://github.com/mapbox/maki/blob/master/LICENSE.txt)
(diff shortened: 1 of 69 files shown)


nrenner (commented):
Nice improvement, thanks!

I'll see what to do about those Font Awesome 5 references.

nrenner:
> 2. the id-tagging-schema references some icons from font-awesome 5. Yet we're using version 4.7 currently, which is missing most of the icons. For the moment those fas-xxx icons are not handled, i.e. the fallback icon (fa-search) is used.

See #398

PR #411, opened 2021-05-12
add "supermarket" POI layer

mjaschen:
This pull request adds an Overpass API POI layer for the OSM tag `supermarket`.


grizzee:
Thanks! Would be nice to know when i can refill the pockets while cycling. 👍

mjaschen:
(closed because based on wrong starting point, I'll will create a new pull request)

PR #414, opened 2021-05-13
Add tooltip with POI name to POI icons

mjaschen:
This pull request adds a tooltip to custom POI icons. The tooltip just shows the name of the POI.


nrenner:
Thanks. Would be useful for Overpass markers, too. There is also [L.Tooltip](https://leafletjs.com/reference-1.7.1.html#tooltip), which would show more instantly (I think).

mjaschen:
I'll take a look.

PR #415, opened 2021-05-14
Replace native HTML tooltip for custom POIs with Leaflet Tooltip

mjaschen:
- the tooltip shows instantly when hovering a POI icon with the mouse cursor
- the content is the same as for the popup, except for the action button (“delete”) as the button isn't reachable by the mouse (the tooltip disappears when the mouse cursor leaves the icon area)
- (maybe we should add a hint to the tooltip, e.g. “Click icon for actions” (*TBD*))
- the tooltip is only attached to POI icons on devices without a touch interface, i. e. when `BR.Browser.touch` is `false`
- the tooltip is removed when the icon is clicked, otherwise tooltip and popup would be visible at the same time
- the tooltip is enabled again after the popup was closed

(I found no obvious solution for showing tooltips on the Overpass Layer icons yet)

### DIFF
--- a/js/plugin/POIMarkers.js
+++ b/js/plugin/POIMarkers.js
@@ -97,23 +97,37 @@ BR.PoiMarkers = L.Control.extend({
             markerColor: BR.conf.markerColors.poi,
         });
 
-        var content = '<p>' + BR.Util.sanitizeHTMLContent(name) + '</p>';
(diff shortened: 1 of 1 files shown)


nrenner (commented):
Nice improvement, thanks.

PR #417, opened 2021-05-16
Use mapbox/polyline for polyline/polygon URI param encoding

### LINKED ISSUE #220: NS_ERROR_MALFORMED_URI when importing nogos with buffering

bagage:
Closes #220 .

Continuation of #233 

Still some issues:

- it blinks when adding a waypoint
- this needs some work on brouter side by either accept encoded param, or use a `POST` instead?

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,15 @@
             "webRoot": "${workspaceFolder}"
         },
         {
+            "name": "Launch Firefox against localhost",
+            "type": "firefox",
+            "reAttach": true,
+            "request": "launch",
+            "profile": "default-release",
+            "url": "http://localhost:3000",
+            "webRoot": "${workspaceFolder}"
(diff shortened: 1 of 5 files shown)


bagage:
Closing as stalled for too long

Issue #402, opened 2021-04-25
Add Swisstopo (Swiss Map Raster or Vector) baselayer


fidelius-dc:
Hello,

As all Swisstopo maps are freely available now, it would be great to add them to the layer selection. In WMS they were alredy available for a long time. It would be a huge advantage for planning in Switzerland.

https://www.swisstopo.admin.ch/en/swisstopo/free-geodata.html#189_1614237763665

nrenner:
The link is more about downloading raw geo data? What do you have in mind specifically?

The only layer types we support right now, is the OSM png tile schema or compatible WMTS layers.

fidelius-dc:
Ok, I see. brouter uses wmts, and I figured out that i need to add

https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.pixelkarte-farbe/default/current/3857/{z}/{x}/{y}.jpeg

as custom baselayer. THANKS.

I think it would be a good idea to add this to the extended list of baselayers for Switzerland.

Further documentation on the swisstopo WMTS here: http://api3.geo.admin.ch/services/sdiservices.html#wmts
Thanks a lot!

phylor:
For all googlers: the following link can be used to add aerial photographs from Swisstopo. Quite useful, e.g. to figure out what type of road it is.

https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swissimage/default/current/3857/{z}/{x}/{y}.jpeg

nrenner:
> https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.pixelkarte-farbe/default/current/3857/{z}/{x}/{y}.jpeg

Generally I'm a bit unsure about adding non-OSM maps, as it might cause confusion when the route and the map don't match and it is a bit at odds with promoting OSM.

But there seems to be a demand for it, so I don't want to be too restrictive either.

fidelius-dc:
Hi, I fully understand your concerns - Brouter made me an OSM contributor in the first place! 

But as it's hidden in the additional layers I consider it as the same like TopPlusOpen. It's rather for the advanced user, that sould be able to figure out, that routing is based on OSM and understand that you'll run into inconsistencies using these maps.

Thanks you

fidelius-dc:
as there is #422 now, I guess we can close this

PR #422, opened 2021-05-23
Add SwissTopo tile layers

phylor:
This PR adds two Swiss tile maps from Swisstopo (the official Swiss street map and aerial photographs). Both maps are freely available since March 2021 and can be retrieved directly from Swisstopo servers.

It has been suggested to add them to the standard map set in #402.

image
image

### DIFF
--- a/layers/config/overrides.js
+++ b/layers/config/overrides.js
@@ -137,6 +137,14 @@ BR.confLayers.getPropertyOverrides = function() {
             'mapUrl': 'https://osm.ch/#{zoom}/{lat}/{lon}',
             'worldTiles': true
         },
+        'swisstopo-landeskarte': {
+            'country_code': 'CH',
+            'mapUrl': 'https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.pixelkarte-farbe/default/current/3857/{z}/{x}/{y}.jpeg',
+        },
+        'swisstopo-aerial': {
+            'country_code': 'CH',
+            'mapUrl': 'https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swissimage/default/current/3857/{z}/{x}/{y}.jpeg',
+        },
         'topplus-open': {
             'country_code': 'DE',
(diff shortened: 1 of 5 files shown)


layers/config/overrides.js
```diff
@@ -137,6 +137,14 @@ BR.confLayers.getPropertyOverrides = function() {
             'mapUrl': 'https://osm.ch/#{zoom}/{lat}/{lon}',
             'worldTiles': true
         },
+        'swisstopo-landeskarte': {
+            'country_code': 'CH',
+            'mapUrl': 'https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.pixelkarte-farbe/default/current/3857/{z}/{x}/{y}.jpeg',
```
nrenner, line 142: `mapUrl` is not the tile url, but a link to the original map that is shown for the active layers above the attribution on the bottom right.

In this case that would probably be https://map.geo.admin.ch/. The current map view can be passed with `{zoom}/{lat}/{lon}` placeholders, but I don't know if that map understands lat/lon.

nrenner (changes requested):

layers/config/overrides.js
fidelius-dc: The map itself certainly understands lat/lon, but not sure it can be adressed like that by URL. By default it works with the "CH1903 LV95" coordinate system. By replacing the &E= and &N= heading by &lat= &lon= it switches to 3D mode.

phylor: According to the [help section](https://help.geo.admin.ch/?ids=54&lang=de), we can use `swisssearch`:

https://map.geo.admin.ch/?swisssearch=47.378319,8.542399

It seems that the `zoom` parameter is ignored though, when using `swisssearch`.

Will update the URL, thank you for the hints!

nrenner (approved):

nrenner:
I added passing the `bgLayer` parameter.

phylor:
@nrenner Thank you very much! :+1:

PR #425, opened 2021-06-02
update license information for BRouter

mjaschen:
BRouter's license was changed from GPL to MIT two years ago. This pull request ensures that the correct license is mentioned in the README.


nrenner (commented):
Good catch, thanks.

PR #456, opened 2021-11-14
remember and recall last used (non-custom) profile

### LINKED ISSUE #437: When revisiting brouter, layers et al. are stored, but not the bike type
I noticed, that layers and so on are stored, however not the type of the bicycle, The memo function (cookie? local storage?) should be harmonised and store all user-changes options/parameters.

mjaschen:
This pull requests enables remembering and recalling the last used non-custom profile.

The profile identifier is written into the local storage when it's activated. 

The stored profile will be activated if found in local storage when the BRouter-web instance is initizialized.

Fixes #437.

### DIFF
--- a/js/control/RoutingOptions.js
+++ b/js/control/RoutingOptions.js
@@ -8,6 +8,9 @@ BR.RoutingOptions = L.Evented.extend({
     initialize: function () {
         $('#profile-alternative').on('changed.bs.select', this._getChangeHandler());
 
+        var remembered_profile = this.getRememberedProfile();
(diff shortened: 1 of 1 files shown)


bagage:
Nice! Thanks @mjaschen for this nice improvement. Looks good to me, waiting for @nrenner before merging it though :)

nrenner (approved):
I haven't tested it, but looks good to me, too. Thanks for the contribution.

@bagage As you mentioned it, do you want to merge?

PR #458, opened 2021-11-17
Add OpenStreetMap Notes Layer

mjaschen:
This commit adds a new layer which renders OpenStreetMap notes in the
current view.

The layer can be activated via the "More" button in the layers sidebar.


nrenner (commented):
Thanks.

PR #461, opened 2021-11-21
Update Docker configs

mjaschen:
This commit updates the Dockerfile so that the image is built on top of the nodejs Long-term support (LTS) version. Manually bumping the nodejs version is no longer required.

Additionally, the ports definition in docker-compose.yml is quoted as string, as suggested in the reference documentation.


bagage:
Thanks @mjaschen :)

PR #464, opened 2021-11-27
fix swisstopo-landeskarte id

sylvainavelo:
Hi,

There is a typo in the id of the swisstopo-landeskarte layer added in PR #422 and it leads to behaviors like not being able to remove the layer once added to baselayers.

(Tell me if you'd rather me have open an issue first)


nrenner (commented):
Thanks!

Issue #494, opened 2022-05-06
Ways with different `surface` tag variants counted twice in track statistics


stesie:
With #460 brouter-web now supports ways with a `cycleway:surface` tag. However, if a way has both a `surface` and a `cycleway:surface` tag (and brouter is invoked with `processUnusedTags` so both are returned), ... then brouter-web counts this way twice in the track statistics:

<img width="262" alt="Screenshot of track statistics w/ ~150m way type and ~310m surface type" src="https://user-images.githubusercontent.com/113068/167133728-8c3ff960-da8d-4dd4-98ad-08d5c14ce543.png">

Example: https://bikerouter.de/#map=19/49.79318/9.94119/osm-mapnik-german_style&lonlats=9.940186,49.793385;9.942198,49.792959

mjaschen:
That's on me I guess ;-)

I'll fix it ASAP.

stesie:
If you've got a brouter backend with `footway:surface` in `lookups.dat` it could even be three to choose from :-)

Prefering any specialized variant should likely be fine unless `processUnusedTags` is set. Then it's up to the profile if it's considering it (since it's a cycling profile) or not. With `processUnusedTags` set however it'll yield wrong statistics with non-bike profiles.

Personally I'd lean towards prefering any specialized variant (and falling back to plain e.g. `surface`) in favour of little complexity and accept wrongs stats for car profiles... (yet, as a cyclist, I might be biased :joy:)

mjaschen:
I think we can safely assume that the majority of BRouter-web users utilize it to plan cycling routes ;-)

So I'm with you and think it's ok to prefer the specialized `*:surface` variants over `surface` when both are present.

mjaschen:
FYI, I've implemented a fix.

I'll wait with the pull request until the branches for #68 are merged.

nrenner:
Could you please make a PR now, planning a release in the coming week.

nrenner:
Fixed in #550

PR #550, opened 2022-05-30
fix for #494

mjaschen:
If a way segment contains more than one `surface` tag (e.g. `surface=*` and `cycleway:surface=*`), the tag best suited for the current routing type (currently only *cycling* is supported) is selected.

i.e. if a `cycleway:surface=` tag belongs to the current way segment it will be used for the analysis exclusively (all other `surface` tags are skipped from now on).

### DIFF
--- a/js/control/TrackAnalysis.js
+++ b/js/control/TrackAnalysis.js
@@ -122,7 +122,7 @@ BR.TrackAnalysis = L.Class.extend({
      * @returns {Object}
      */
     calcStats: function (polyline, segments) {
-        var analysis = {
+        const analysis = {
             highway: {},
             surface: {},
             smoothness: {},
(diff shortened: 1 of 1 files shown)


nrenner:
Hm, just tested afterwards and it always shows `undefined` (?):

`normalizedWayTags` is defined as an array,
https://github.com/nrenner/brouter-web/blob/af1bfc224f3591ff2193a7047e81efd9631d2824/js/control/TrackAnalysis.js#L209

but used as an object (adds properties to the Array object, but does not push to the array)
https://github.com/nrenner/brouter-web/blob/af1bfc224f3591ff2193a7047e81efd9631d2824/js/control/TrackAnalysis.js#L237

so the length of the array is 0 here and further processing is skipped:
https://github.com/nrenner/brouter-web/blob/af1bfc224f3591ff2193a7047e81efd9631d2824/js/control/TrackAnalysis.js#L144

PR #553, opened 2022-06-06
Fix for incomplete/buggy solution for #494

mjaschen:
This commit fixes the incomplete/buggy implementation of the fix for #494.
Values are now converted to the expected format before returned from
`wayTagsNormalize()`.


nrenner (commented):
Looks good to me now, thanks.

Issue #467, opened 2021-12-13
OverpassLayer: include OSM link in pop-up


Dirk--:
nice to have: Link to OSM object (for futher investigation like history, dropping a note etc.)

brouter-web/js/LayersConfig.js > createOverpassLayer > 

`title: '{{ type }} <a href="https://www.openstreetmap.org/{{ type }}/{{ osm_id }}" target="_blank">{{ osm_id }}</a>',`

example:
https://brouter.grade.de/#map=19/53.09885/5.94807/CARTO,lock,lock-gate

Dirk--:
an other on on my wish list: 
_generic overpass query_ https://github.com/plepe/overpass-layer/issues/12
(like the 'question mark functionality' at lower right side of https://www.openstreetmap.org)

plepe:
Check this out: https://github.com/plepe/leaflet-overpass-lens

plepe:
I added some examples how to use the module in the README file. Also, I added a link to a short screencast: https://www.youtube.com/watch?v=JTzVa9E0Yo8

I hope, you find this useful!

mjaschen:
I've submitted a PR which adds a “Inspect at OSM” link to the Overpass popup (see #574).

PR #574, opened 2022-06-30
Improves rendering of Overpass data; add OSM link to Overpass popups

mjaschen:
(see #467)

- add some styling to the Overpass data table (alternating row colors; text alignment)
- add link to OSM object below the Overpass data table
- add 4 new entries in `en.json`

### DIFF
--- a/js/LayersConfig.js
+++ b/js/LayersConfig.js
@@ -216,7 +216,7 @@ BR.LayersConfig = L.Class.extend({
                 minZoom: 12,
                 feature: {
                     title: '{{ tags.name }}',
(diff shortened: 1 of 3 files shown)


js/LayersConfig.js
```diff
@@ -240,6 +240,61 @@ BR.LayersConfig = L.Class.extend({
...
+            '</th></tr>';
+        output += '</thead>';
+
+        output += '<tbody>';
+        for (const key in overpassData.tags) {
+            if (key.substring(0, 5) === 'addr:') {
+                continue;
+            }
+            // `new Option().innerHTML` escapes HTML entities for XSS protection
+            let value = new Option(overpassData.tags[key]).innerHTML;
```
nrenner, line 263: Some afterthoughts on sanitizing, I can do a commit, but wanted to ask:
- we have [Util.sanitizeHTMLContent](https://github.com/nrenner/brouter-web/blob/f68d32b1f2b5ea34202c2bd62d7a91770942b7f0/js/Util.js#L137), is there a difference? Should we use that (current short enough to leave as is)?
- I would also sanitize the key, OSM API doesn't restrict allowed characters, see e.g. [TagInfo Characters in keys](https://taginfo.openstreetmap.org/reports/characters_in_keys#problem)

Issue #310, opened 2020-06-08
Color-coding the elevation profile by track type [request for comments]


stesie:
This is more like a request for comments (for the moment) ...

Inspired by @mjaschen's work on track type analysis I was wondering whether it would be possible to color-code the elevation profile. The idea essentially is to depict the track type (e.g. track/path/other) relative to current gradient -- like you might not want to have paths uphill and grade1 tracks downhill.

For the moment I've just hacked away and this is how the proof of concept result looks like:

![color-coded-elevation-chart](https://user-images.githubusercontent.com/113068/84080279-3aa4ce80-a9dc-11ea-9ffa-ffabec5aeb43.png)

(red = path, yellow = track, red = other)

What do you think about it?

For the moment I'm wondering how (i.e. what color for which track type - and maybe what (surface?)) to color code. If you've got ideas or recommendations, please let me know.

Implementation-wise one drawback is that BR.Elevation plugin and leaflet-elevation control need to be coupled tighter. However leaflet-elevation is a fork already anyways (so not modified much yet) and it has a generic GPX/GeoJSON interface (i.e. only latlngs are passed) ... but not a "control every single stroke" kind of one.
I'm uncertain what to do about that - especially since I'm all new to GeoJSON et al, if there's some generic interface that could be used ... or whether to just let go and copy the file over to brouter-web!?

Looking forward to your feedback

nrenner:
My idea was to have such color coding as a separate bar below the elevation profile. 

I have seen this somewhere, but can't remember where, will need to check some sites. 

Such a bar would be wider and better recognizable than the thin outline. And I find coloring the elevation outline distracts and makes it harder to see the actual elevation shape.

I find your suggestion with track/path/other a bit too MTB-specific, ideally it would be more generic and one of multiple options. The overall goal would be that the available options for coloring and color codes are the same as for - and maybe in sync with - the track analysis (#304) and the route quality coding (#242, with the suggested extension by MultiOptionsPolyline to also support the analysis types).

Regarding the implementation, my theoretical philosophy is to use as much libraries and plugins as possible, to keep them external and ideally contribute back. In practice, we haven't always followed this and also copied two plugins over, so this is not a strict policy and I don't want to set a too high bar for contributions.

For leaflet-elevation specifically I was thinking about merging some of the pending pull requests and maybe suggesting to create a Leaflet GitHub organization for commonly maintaining abandoned plugins like this.

Ideas for adding coloring might be:
[...]

PR #621, opened 2022-08-28
Feature Request: Road surface as quality color

tbsmark86:
This is similar to #310

Of course the actual numbers used are highly subjective.

Notes:
* Required fork of leaflet-hotline with a new option to disable the gradient display.
  This is was done so that short stretches of very bad surface adjacent to very good ones are visible.
  (The original author does not maintain it anymore so no harm in forking.)
* I've ignored the compat warning:
    URLSearchParams is not supported in Safari 7, op_mini all, IE 10, android 4.1
   don't seem relevant today. Those are all EOL maybe eslint configuration should be updated?
   And supporting opera-mini seems pointless.

### DIFF
--- a/js/plugin/RoutingPathQuality.js
+++ b/js/plugin/RoutingPathQuality.js
@@ -60,6 +60,159 @@ BR.RoutingPathQuality = L.Control.extend({
                     },
                 }),
             },
+            surface: {
+                title: i18next.t('map.route-quality-surface'),
+                icon: 'fa-road',
+                provider: new HotLineQualityProvider({
+                    hotlineOptions: {
+                        renderer: renderer,
+                        palette: {
+                            // normal range
+                            0.0: '#ff0000',
+                            0.95: '#00ff00',
+                            // special value for unknown
+                            1.0: '#888888',
+                        },
+                        // note: without this the lib will get min/max from the actual
+                        // values rendering the special values moot
+                        min: 0,
+                        max: 1,
+                        discreteStrokes: true,
+                    },
+                    valueFunction: (function () {
+                        let cache = [];
+                        return function (latLng) {
+                            var feature = latLng.feature;
+                            if (!feature.wayTags) {
+                                return 1.0;
+                            } else if (cache[feature.wayTags]) {
+                                return cache[feature.wayTags];
+                            }
+                            let data = new URLSearchParams(feature.wayTags.replace(/\s+/g, '&')); // eslint-disable-line compat/compat
+                            let surface = null;
+                            switch (data.get('surface')) {
+                                case 'paved':
+                                    surface = 0.8;
+                                    break;
+                                case 'asphalt':
+                                case 'concrete':
+                                    surface = 1;
+                                    break;
+                                case 'concrete:lanes':
+                                case 'concrete:plates':
+                                    surface = 0.6;
+                                case 'sett':
+                                case 'gravel':
+                                case 'pebblestone':
+                                    surface = 0.5;
+                                    break;
+                                case 'paving_stones':
+                                case 'compacted':
+                                case 'fine_gravel':
+                                    surface = 0.7;
+                                    break;
+                                case 'cobblestone':
+                                case 'dirt':
+                                case 'grass':
+                                    surface = 0.2;
+                                    break;
+                                case 'unhewn_cobblestone':
+                                    surface = 0.01;
+                                    break;
+                                case 'ground':
+                                case 'earth':
+                                    surface = 0.3;
+                                    break;
+                                case 'mud':
+                                case 'sand':
+                                    surface = 0.01;
+                                    break;
+                                case null:
+                                    break;
+                                default:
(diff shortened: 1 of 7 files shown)


js/plugin/RoutingPathQuality.js
```diff
@@ -60,6 +60,159 @@ BR.RoutingPathQuality = L.Control.extend({
...
+
+                            // modifier tracktype; also sometimes only tracktype is available
+                            if (data.get('highway') === 'track') {
+                                switch (data.get('tracktype') || 'unknown') {
+                                    case 'grade1':
+                                        if (surface === null) {
+                                            surface = 0.9;
+                                        } /* else {
+					    don't change
+					} */
```
nrenner, line 145: Tabs instead of spaces

js/plugin/RoutingPathQuality.js
```diff
@@ -60,6 +60,159 @@ BR.RoutingPathQuality = L.Control.extend({
...
+                                    surface = 0.3;
+                                    break;
+                                case 'mud':
+                                case 'sand':
+                                    surface = 0.01;
+                                    break;
+                                case null:
+                                    break;
+                                default:
+                                    console.warn('unhandled surface type', data.get('surface'));
```
nrenner, line 133: Please comment out console warning, as with a selected choice of values there will always be warnings without any actionable implication.

js/plugin/RoutingPathQuality.js
```diff
@@ -60,6 +60,159 @@ BR.RoutingPathQuality = L.Control.extend({
...
+                                        surface *= 0.2;
+                                        break;
+                                    case 'impassable':
+                                        surface *= 0.01;
+                                        break;
+                                }
+                            }
+
+                            // limit normal values 0-0.9 so 1.0 can be unknown
+                            const final = surface === null ? 1.0 : surface * 0.9;
```
nrenner, line 209: `surface=asphalt` + `smoothness=excellent` looks more like unknown, probably because `final` = ~0.99 (1 * 1.1 * 0.9), which is closer to gray for unknown than green, example:
- http://localhost:3000/#map=16/47.6917/9.0687/standard,route-quality&lonlats=9.062669,47.692158;9.070222,47.690237
- [way ‪147345349‬](https://www.openstreetmap.org/way/147345349)

tbsmark86:
To solve this "what color is what" a bit better I added a more usable way to find the message for a certain part of the route.

> What I also still have in mind is separate color codings of highway, surface and smoothness tag values.

Yeah sounds nice. But I feel there are to many different surface values ... so finding colors that are somewhat self-explanatory could be very hard. For example is light-grey concrete or cobblestone?
This soulution is at least something. You could also rename it 'smoothness' rating because essentially red is very rough (dirt, cobbelstone) and green is very smooth (asphalt) - so no real conflict :)

> I personally don't like the simple red-green color palette

Ok. I did put much thought into the colors.

nrenner (changes requested):
> To solve this "what color is what" a bit better I added a more usable way to find the message for a certain part of the route.

Nice, but just finding the closest node will sometimes return a node from another road nearby. See this example, when hovering along the "Europastraße" (highway=primary) it will also highlight the two other segments in some parts:

http://localhost:3000/#map=19/47.66680/9.16253/standard&lonlats=9.162721,47.666876;9.162458,47.667085

The height marker jumps to the other segments as well, it looks like you've actually taken the [code from Leaflet.Heightgraph](https://github.com/GIScience/Leaflet.Heightgraph/blob/9e45353ae01e4389124701bdacb62ee255d3fec1/src/L.Control.Heightgraph.js#L1041-L1063).

A better solution might be to use [Turf.nearestPointOnLine](http://turfjs.org/docs/#nearestPointOnLine), then maybe the `index` or `location` return values can help finding the next/previous node along the line (again not closest). There are also [Turf.pointToLineDistance](http://turfjs.org/docs/#pointToLineDistance), [L.LineUtil.closestPointOnSegment](https://leafletjs.com/reference.html#lineutil-closestpointonsegment) and [L.LineUtil.pointToSegmentDistance](https://leafletjs.com/reference.html#lineutil-pointtosegmentdistance).

> You could also rename it

Yes, my comment was just about the naming, but that can be addressed once there is actually a conflict. Both approaches are valid and not exclusive.

tbsmark86:
Added the proposed changes. Less code & more reuse nice! 
Yes, I've used the code from  Heightgraph which of course means that this has the same bug. But probably less relevant there.

Here the most extreme example for Heightgraph I could find (which also led me to another bug in my code ...):
[http://localhost:3000/#map=17/46.59409/11.52860/standard&lonlats=11.527732,46.594296;11.527188,46.593598&profile=fastbike](url)

nrenner (approved):
> Yes, I've used the code from Heightgraph which of course means that this has the same bug. But probably less relevant there.

Yes, I would also consider it undesired behavior there, but as the marker and info is shown at the selected point it's less of an issue.

Thanks for your contribution!

PR #630, opened 2022-09-11
Add OpenRailwayMap

emesik:
This is a useful overlay extracting railway data.


nrenner:
Thanks for your contribution!

JOSM layers can be added and updated with the [extract.mjs](https://github.com/nrenner/brouter-web/blob/master/layers/josm/extract.mjs) script using `yarn layers` and the ID [defined](https://josm.openstreetmap.de/wiki/Maps/Worldwide#OpenRailwayMap-Default) in the JOSM Maps Wiki (I need to document that). So I added a second commit switching to that.

Issue #463, opened 2021-11-27
Add IGN France baselayers and overlays


sylvainavelo:
Hello,

Since summer 2021, IGN France web services are accessible without key (as it used to be for years).
Lists in french : https://geoservices.ign.fr/services-web-experts

I suggest to add baselayers such as aerial pictures, a base map and an overlay with administrative boundaries. All of those are WMTS tile layers compatible with brouter map display.

The idea comes from a french cyclist requesting a way to add french departments boundaries on BRouter, and using an IGN WMTS layer is a possible solution : https://twitter.com/sylvainavelo/status/1462853551139524615

It will be very much like what has been done for Swisstopo baselayers (#402).
I've seen and I understand the concern about adding non-OSM maps.

I'm working on the issue but will submit the PR only if it meets approval.

nrenner:
What are the usage terms though? E.g. in Germany official maps are often open for private use but not for embedding in map services.

I don't see the use case or general demand for admin boundaries, so I would rather not want those.

sylvainavelo:
Usage is free for any use, including commercial, under the terms of this license : https://www.etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf
Like a CC-BY, mentioning paternity is required.
And also the date of last update... But as it is a web service, I can't figure out how this can be achieved. It's probably a minor concern.

The license is clear for the data itself but it lacks information specific to web services. How much can you request the service ? (in terms of requests per minutes). I can't find. The Swiss give that information quite clearly [here](https://www.geo.admin.ch/de/geo-dienstleistungen/geodienste/terms-of-use.html) for instance. I will ask for details.

nrenner:
> I will ask for details.

Have you asked?

Not sure that would help much, as we can't really tell or control what our usage would be.

sylvainavelo:
Yes I asked and it took some time to have this answer. Sorry for the late follow up :

- They contact users individually when they have clues on how to find them if their usage is dangerous for their infrastructure.
- They block IPs if necessary but not automatically.
- On their current infrastructure, it is time consuming to track users that have an unfair usage.
- They are aware that in most cases, the client application has no mean to track unfair usage as well.
- Use in an application with 10 000 users a day would be fair use (I am surprised they said users per day and not a metric using the number of requests). That seems not very specific.

nrenner:
> Use in an application with 10 000 users a day would be fair use 

[...]

PR #636, opened 2022-09-12
feat/ignf aerial pictures and base map

sylvainavelo:
Adds IGN France base layers : aerial pictures and base map

These layers do not require a key.

----

Related issue : #463


nrenner (commented):
Thanks for your contribution!

I added a commit to assign the `franceBbbox` to the layer `geometry` property. It's shown in the layer preview but does not restrict the layer bounds with `worldTiles` set, because of the world wide coverage in lower zooms.

Issue #322, opened 2020-07-06
Exporting to ANT+/Garmin FIT


cryptomilk:
Would it be possible to support to export in FIT file format?

The FIT documentation is included in the FIT SDK archive (http://www.thisisant.com/resources/fit/).

rkflx:
@cryptomilk: If you don't mind the question: What's your use case for exporting FIT files?

Did Garmin scrape GPX import from their newer devices? Is there an app which only takes FIT files, but no GPX?

(Importing FIT might also be worth considering, though, to avoid having to convert FIT to GPX with GPSBabel before viewing recorded tracks.)

cryptomilk:
I have a Garmin Fenix Smartwatch, if you only import a GPX, the track is displayed on the map, but thats it..

With a FIT file from Garmin you get turn instructions. The watch e.g. informs you that in 100m there is a turn left or right. I think this is part of the FIT file format but GPX doesn't have such an extension to store such information.

I guess it is the same for other Garmin devices.

cryptomilk:
So currently I create a GPX track with Brouter, save it, import it in Garmin Connect and send it to the device. I would love if I would not have to use Garmin Connect but directly put it on my watch.

nrenner:
I'm generally unsure about the way forward with formatting.

The currently supported formats are own implementations in the server and I personally would be a bit reluctant to implement more this way. I haven't found a JavaScript GPX to FIT converter on a quick search though.

The GPX formatter supports various turn instruction styles, as every app seems to have their own, see [OsmTrack.java](https://github.com/abrensch/brouter/blob/0841216cda47d3327b039e6db00c253e86d0948d/brouter-core/src/main/java/btools/router/OsmTrack.java#L403) (search for `turnInstructionMode`).

Are you sure that Garmin doesn't support turn instructions in GPX?

rkflx:
> I guess it is the same for other Garmin devices.

AFAIK for Garmin Edge devices (i.e. cycling), you've got the following options for a free-software-only workflow:

A) Import track as GPX file, enable turn instructions on the device's course selection screen.
  - Calculated on-device
  - Rerouting possible (e.g. after a missed turn)
  - Turns are shown embedded in the map (exactly where they belong) and additionally as a pop up dialog with a zoomed-in map

B) Use another app or website (e.g. https://cycle.travel) to create a TCX file, and make sure to select the option to embed turn instructions as "course points".
  - Calculated in advance by the respective app or website
[...]

PR #641, opened 2022-09-13
 Exporting to FIT #322

tbsmark86:
I've finally (after some failed runs ...) had a successful test ride (9km, 11 turns) with correct turn instructions on my wahoo bolt (old version) today.

* Tip: You can use https://www.fitfileviewer.com/  to view the content
* The 3 patch is not required for this I've just stumbled upon this problem while searching for the distance value of the turn markers ... before realizing this is distance-to-next and not the distance-from-last that I needed :)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,5 @@
 module.exports = {
     presets: [['@babel/preset-env', {}]],
     sourceType: 'script',
-    exclude: [/node_modules\/(?!overpass-layer|leaflet.locatecontrol\/).*/],
+    exclude: [/node_modules\/(?!overpass-layer|leaflet.locatecontrol|fit-file-writer\/).*/],
 };
--- a/css/style.css
+++ b/css/style.css
@@ -590,6 +590,10 @@ table.dataTable.display tbody tr:hover.selected {
     line-height: 24px;
 }
 
+.format-turns-enabled {
+    font-size: 0.8em;
+}
+
 /* tooltip */
 
 .editing-tooltip,
--- a/index.html
+++ b/index.html
@@ -450,6 +450,7 @@
                                                     checked
                                                 />
                                                 <span data-i18n="export.format_gpx">GPX</span>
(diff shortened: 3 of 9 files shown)


tbsmark86:
About this build error: gulp uglify complains about the es6 code in the ne fit-file-writer lib.
However if i move the code from node_modules/... to js/export/ it does not complain??

I've never used gulp before any tips? Or do I have to rewrite the lib to use es5?

tbsmark86:
force push to fix the build error

nrenner (changes requested):
Thanks for the PR!

> The 3 patch is not required for this I've just stumbled upon this problem while searching for the distance value of the turn markers ... before realizing this is distance-to-next and not the distance-from-last that I needed :)

Thanks for the fix. Your "distance-from-last" comment got me confused, as I understand this fixes distance-to-end-of-segment instead of distance-to-next?

However, my primary focus for client-side formatting so far was to produce largely the same results as the brouter server/app. So `voicehints` of the expected test result [brouterTotal.json](https://raw.githubusercontent.com/nrenner/brouter-web/master/tests/control/data/brouterTotal.json) should match the [total download](https://brouter.de/brouter?lonlats=8.467712,49.488117|8.469354,49.488394|8.470556,49.488946|8.469982,49.489176&profile=trekking&alternativeidx=0&format=geojson&profile:turnInstructionMode=2) from the server (for [this route](https://brouter.de/brouter-web/#map=18/49.48856/8.46986/standard&lonlats=8.467712,49.488117;8.469354,49.488394;8.470556,49.488946;8.469982,49.489176)).

I would appreciate if you could move the third commit 54bf446 to a separate PR, as I'm not sure if we should wait until the pending server-side PR is merged that supposedly also fixes this.

tbsmark86:
Ok. I removed the third commit - extra pull request incoming soon.

nrenner (approved):

nrenner:
Thanks for your contribution!

PR #642, opened 2022-09-16
Add Strava heatmaps

emesik:
These have max zoom level of 12, so I'm not sure if you want to include them.

### DIFF
--- a/layers/config/overrides.js
+++ b/layers/config/overrides.js
@@ -216,6 +216,12 @@ BR.confLayers.getPropertyOverrides = function() {
             'nameShort': 'MTB',
             'mapUrl': 'http://mtb.waymarkedtrails.org/#?map={zoom}!{lat}!{lon}'
         },
+        'strava_cycling_heatmap': {
+            'overlay': true
+        },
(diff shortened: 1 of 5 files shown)


nrenner:
I'm not sure if the heatmap is supposed to be used by third-party apps?

Where did you get the URL? What are the terms of use?

emesik:
I've taken it from JOSM, the way you described it in #630

nrenner (commented):
Right, I hadn't seen that. Still, many editor layers, like imagery, are licensed for OSM editing only.

And [according to the wiki](https://wiki.openstreetmap.org/wiki/Strava#Data_Permission_-_Allowed_for_tracing!) that also applies to the Strava heatmap:
> OpenStreetMap users have permission to use Strava heatmap data for tracing into OSM only. For all other non-personal uses, [...] please contact Strava Metro.


So unless there is an official page explaining how to use those tiles and under what terms, e.g. on developers.strava.com (haven't found one on a quick search), I don't think we can use the heatmap, unfortunately.

Thanks for the PR, anyway.

PR #646, opened 2022-09-26
Turn instructions had wrong distance when using multiple segments

tbsmark86:
Split of from pull-request #641 

The distance from the start for the a segment till the next turn is missing from the distance value of the last turn in the previous segment.
However it's not really that important right now because it's never visible to the user and only used when exporting to gpx in locus-style. (locus:rteDistance and locus:rteSpeed).

> Your "distance-from-last" comment got me confused,

Yeah it's all really confusing :)

I've added more explanation to the code maybe this clears it up - or make it worse.

> So voicehints of the expected test result brouterTotal.json

Ok. I've changed the test so the brouterTotal.json remains untouched.

> should match the [total download](https://brouter.de/brouter?lonlats=8.467712,49.488117%7C8.469354,49.488394%7C8.470556,49.488946%7C8.469982,49.489176&profile=trekking&alternativeidx=0&format=geojson&profile:turnInstructionMode=2) from the server (for [this route](https://brouter.de/brouter-web/#map=18/49.48856/8.46986/standard&lonlats=8.467712,49.488117;8.469354,49.488394;8.470556,49.488946;8.469982,49.489176)).

Ah ok. I did not know that the server could handle multiple segments. So yes you're right it's also a backend bug. Consider [this route without the first waypoint](https://brouter.de/brouter?lonlats=8.467712,49.488117%7C8.470556,49.488946%7C8.469982,49.489176&profile=trekking&alternativeidx=0&format=geojson&profile:turnInstructionMode=2):
The distance for the first turn changes from 88 to 140. That's the almost the same as my calculated value of 141.

Is this important enough to also open a bug for brouter itself?  Because unless someone is using the old export it should not affect a brouter-web user once this patch is merged?

### DIFF
--- a/js/control/Export.js
+++ b/js/control/Export.js
@@ -322,17 +322,39 @@ BR.Export._concatTotalTrack = function (segments) {
             }
             if (fp.voicehints) {
                 if (!p.voicehints) p.voicehints = [];
+                let prevHintIdx = p.voicehints.length - 1;
+                let hintDistanceSum = 0;
                 for (const fpHint of fp.voicehints) {
                     const hint = fpHint.slice();
                     hint[0] += coordOffset;
+                    hintDistanceSum += hint[3];
                     p.voicehints.push(hint);
                 }
+                // fix the distance for the last turn in a segment
+                //
+                // Problem: In a segmented route where each segment is requested
+                // separately the server can't see that we intend to go further.
+                // Therefore the last voicehints/turn can only contain the
+                // distance till the end of that segment.
+                // However: The distance from the start of segment to the next
+                // turn is missing.
+                // Furthermore the server never reports the distance from the
+                // start to the first turn. So we have to calculate that too.
+                if (prevHintIdx >= 0) {
+                    let firstHintDistance = parseInt(fp['track-length'], 10) - hintDistanceSum;
+                    p.voicehints[prevHintIdx][3] += firstHintDistance;
+                }
+            } else if (p.voicehints && p.voicehints.length) {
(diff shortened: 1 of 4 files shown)


nrenner:
Thanks for splitting up and the clarifying explanation.

I think the goal should be that an exported route from brouter-web always produces basically the same result as the brouter server and app.

Therefore I would want to wait for the pending PR [brouter#441](https://github.com/abrensch/brouter/pull/441), that changes segment-based handling to full route.

tbsmark86:
Hi again!

This is now fixed in upstream so it should be ready for merge now?
I've rebased my commit & updated the test export with a more recent brouter. Sadly there are now way more rounding errors then before.

Because the old test data had no time&energy field there is some new difference between client and brouter-app export.
Maybe also fix this in BR.Export._concatTotalTrack() ? 

Don't know why github build is failing :/

As a reminder what this is about:
* https://brouter.de/brouter-web/#map=18/49.48859/8.46885/osm-mapnik-german_style&lonlats=8.467712,49.488117;8.469354,49.488394;8.470556,49.488946;8.469982,49.489176
* change turninstructions to commentStyle
* export both from client and server
* now view the exported "d2next" column
  * client (before this patch) is: 86, 98, 9
  * server is: 138, 98, 58

afischerdev:
@tbsmark86 
Thanks for your nice example, it gave me the chance to test it against the [last library](https://github.com/abrensch/brouter/pull/742). I found a problem in voicehint calculation. Update is now running on test server.

PR #647, opened 2022-09-26
Fix #501 missing altitude data caused js-error in leaflet-hotline

### LINKED ISSUE #501: Some ferry lines(?) freeze the map layer
Some routes in Norway (I guess if they belong to a ferry line where some keys are missing) freeze the map layer. You cannot drag it with the mouse anymore and clicking in it does not work.

An example is this link: https://brouter.m11n.de/#map=12/60.0736/5.4358/CyclOSM,Waymarked_Trails-Cycling,Smoothness,route-quality&lonlats=5.436859,60.161993;5.335922,59.942872&profile=m11n-gravel
Also, the track remains gray.

For ferry connections in denmark I see no problem, also the heightdata of the ferry line is shown.

Reproduced with Firefox and Qutebrowser (QtWebEngine/Blink).

tbsmark86:
Thats the simple solution. Missing values are displayed as zero altitude.

Alternate approach would be extending leaflet-hotline to not draw anything for an missing/NaN value. That shouldn't be to hard if you like it more let me know.

### DIFF
--- a/js/plugin/RoutingPathQuality.js
+++ b/js/plugin/RoutingPathQuality.js
@@ -40,7 +40,7 @@ BR.RoutingPathQuality = L.Control.extend({
                     valueFunction: function (latLng, prevLatLng) {
                         var deltaAltitude = latLng.alt - prevLatLng.alt, // in m
                             distance = prevLatLng.distanceTo(latLng); // in m
-                        if (distance === 0) {
+                        if (distance === 0 || isNaN(deltaAltitude)) {
                             return 0;
                         }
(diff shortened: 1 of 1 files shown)


nrenner:
Thanks for the PR, but I personally find it wrong to assume missing elevations as 0.

I would prefer fixing leaflet-hotline to either leave a gap or draw a distinct color (e.g. gray) for missing values, but that's probably not worth the effort, as [brouter#441](https://github.com/abrensch/brouter/pull/441) will add interpolation and filling up of missing elevations (although I find that equally wrong, except for bridges and tunnels maybe).

PR #672, opened 2022-11-05
Bugfix: Selected edges stick on map after route clear

tbsmark86:
Hover clear is just to be safe very hard to trigger this as a bug.


nrenner (commented):
Thanks!

PR #693, opened 2023-01-10
Improve 'route quality cost' color coding

quaelnix:
This improves the color coding of the 'route quality cost' overlay by omitting 5% of the largest cost values when calculating min-max values:
| Old | New |
| :---: | :---:|
| old | new |

Example: https://brouter.de/brouter-web/#map=11/47.6272/11.3561/Stamen.Terrain,route-quality&lonlats=11.559572,47.760368;11.08792,47.493403&profile=fastbike

### DIFF
--- a/js/plugin/RoutingPathQuality.js
+++ b/js/plugin/RoutingPathQuality.js
@@ -221,6 +221,7 @@ BR.RoutingPathQuality = L.Control.extend({
                 icon: 'fa-usd',
                 provider: new HotLineQualityProvider({
                     hotlineOptions: {
+                        pct: 0.95, // skip (1 - pct) percent of largest values when calculating maximum
                         outlineColor: 'dimgray',
                         renderer: renderer,
                     },
@@ -392,7 +393,8 @@ var HotLineQualityProvider = L.Class.extend({
             if (flatLines.length > 0) {
                 var hotlineOptions = L.extend({}, this.hotlineOptions);
                 if (!hotlineOptions.min && !hotlineOptions.max) {
-                    var minMax = this._calcMinMaxValues(flatLines);
+                    hotlineOptions.pct = hotlineOptions.pct ? hotlineOptions.pct : 1.0;
+                    var minMax = this._calcMinMaxValues(flatLines, hotlineOptions.pct);
                     hotlineOptions.min = minMax.min;
                     hotlineOptions.max = minMax.max;
                 }
@@ -429,14 +431,10 @@ var HotLineQualityProvider = L.Class.extend({
         return [latLng.lat, latLng.lng, val];
     },
 
-    _calcMinMaxValues: function (lines) {
-        var min = lines[0][2],
-            max = min;
-        for (var i = 1; lines && i < lines.length; i++) {
-            var line = lines[i];
-            max = Math.max(max, line[2]);
-            min = Math.min(min, line[2]);
-        }
+    _calcMinMaxValues: function (lines, pct) {
+        lines.sort(function(a, b){return a[2] - b[2]});
+        var min = lines[0][2];
+        var max = lines[Math.ceil(pct * lines.length) - 1][2];
         if (min === max) {
             max = min + 1;
         }


nrenner (commented):
Thanks for the PR!

This looks like an overall improvement. However, in this case, at the start of the route, the segments around the bridge with a base cost/km of 1100 and some turn cost now have the same red color as the prohibitive >= 10000 cost/km pedestrian segments.

Instead of making the top 5% equal (no matter what distribution), an alternative approach might be to use a custom palette, maybe with an exponential step and fixed min/max values like this:

```js
                    hotlineOptions: {
                        ...  
                        palette: { 0.0: 'green', 0.05: 'yellow', 0.1: 'red', 1.0: 'darkred' },
                        min: 1000,
                        max: 10000,
                        discreteStrokes: true,
                        smoothFactor: 0,
                    },
```

These added options might also make sense for the cost coloring:
- `discreteStrokes: true`: turns off gradients as they wash out colors of shorter segments (introduced in [#621](https://github.com/nrenner/brouter-web/pull/621))
- `smoothFactor: 0`: disables line simplification, so short segments (often with turn cost) won't disappear on some zoom levels

| Link | PR max-5% | Palette |
| :---: | :---:| :---:|
| [Start](https://brouter.de/brouter-web/#map=18/47.76052/11.55746/standard,route-quality&lonlats=11.559572,47.760368;11.08792,47.493403&profile=fastbike) | PR start | Palette start |
| [Overview](https://brouter.de/brouter-web/#map=11/47.6309/11.3163/standard,route-quality&lonlats=11.559572,47.760368;11.08792,47.493403&profile=fastbike) | PR overview | Palette overview |

The palette steps are just from a quick trial and error to somewhat match the result of this PR. This is just a rough idea that would need more thought and testing. So I'm not really sure if this would be a better approach?

quaelnix:
> Instead of making the top 5% equal (no matter what distribution), an alternative approach might be to use a custom palette, maybe with an exponential step and fixed min/max values like this

The problem with using fixed min/max values is that there is a wide variety of profiles, and some of them ([including mine](https://github.com/quaelnix/brouter-profiles/blob/main/gravel.brf)) do not strictly adhere to the recommendation to keep the average cost factor of the best route close to one. A viable compromise might be to reserve the dark red shade for the sections where the cost factor is highest.

> as the prohibitive >= 10000 cost/km pedestrian segments.

Segements with a costfactor of 10 are not uncommon in some mountain bike or gravel profiles and do not necessariliy mean that you have to push your bike.

> So I'm not really sure if this would be a better approach?

I like the idea, but I don't think that it is possible to find min/max values that work well with all common profiles.

quaelnix:
> Instead of making the top 5% equal (no matter what distribution), an alternative approach might be to use a custom palette, maybe with an exponential step and fixed min/max values like this

The problem with using fixed min/max values is that there is a wide variety of profiles, and some of them ([including mine](https://github.com/quaelnix/brouter-profiles/blob/main/gravel.brf)) do not strictly adhere to the recommendation to keep the average costfactor of the best route close to one. A viable compromise might be to reserve the dark red shade for the sections where the costfactor is highest.

> as the prohibitive >= 10000 cost/km pedestrian segments.

Segements with a costfactor of 10 are not uncommon in some mountain bike or gravel profiles and do not necessariliy mean that you have to push your bike.

> So I'm not really sure if this would be a better approach?

I like the idea, but I don't think that it is possible to find min/max values that work well with all common profiles.

nrenner:
> The problem with using fixed min/max values is that there is a wide variety of profiles, and some of them ([including mine](https://github.com/quaelnix/brouter-profiles/blob/main/gravel.brf)) do not strictly adhere to the recommendation to keep the average costfactor of the best route close to one.

Might as well work without the fixed min/max. Ideally, the palette steps would be calculated based on the actual cost distribution, but this is getting complicated.

> A viable compromise might be to reserve the dark red shade for the sections where the costfactor is highest.

There could be a separate color assignment for values > max, but that might be difficult to add and probably not something that you would want in the hotline lib itself.

>> as the prohibitive >= 10000 cost/km pedestrian segments.

> Segements with a costfactor of 10 are not uncommon in some mountain bike or gravel profiles and do not necessariliy mean that you have to push your bike.

My bad, I confused cost of 10000 with the "forbidden" cost *factor* of 10000.

I still think, the main issue in the linked example is that the default color palette is linear, while the cost factors in the profile are growing more like exponentially, e.g. some example highway factors from the [fastbike profile](https://github.com/abrensch/brouter/blob/5571eee82c6ea2066af9f6f9208d90950dc66cb6/misc/profiles2/fastbike.brf#L196-L204):
1.0, 1.1, 1.2, 1.3, 5, 10, 1000

But I don't know how to find some kind of exponential palette that matches for all profiles. So the 5% cut might be a practical solution for now, even with the drawback of making higher costs equal.

I can merge the PR as is, or do you have any further work planned?

quaelnix:
> Ideally, the palette steps would be calculated based on the actual cost distribution, but this is getting complicated.

Yes, I totally agree with that, and one of the first things I did when trying to improve the current implementation was to generate a plot of a typical cost distribution of my own profile (the gray vertical dashed line marks the 95th percentile):
segment_cost

> I still think, the main issue in the linked example is that the default color palette is linear, while the cost factors in the profile are growing more like exponentially

I think you're right, but I'm not going to have time in the next few weeks or probably months to develop an implementation that adequately addresses this. So I would appreciate if we could merge this and just keep in mind for later that there is probably room for improvement.

PR #697, opened 2023-02-01
Updated INSTALL.md (mkdir command)

svnkswttr:
added parameter "-p" to the mkdir command to allow creation of parent directories if they not already exist.


nrenner (approved):
Thanks for your contribution!

Issue #666, opened 2022-10-27
Use parameters instead of custom profile upload


zod:
Currently changing profile parameters using brouter-web causes a new custom profile to be uploaded to BRouter. Because custom profiles are only cached and not stored persistently the brouter-web URL doesn't contain a reference of the profile. I sometimes store routes as bookmarks, but this doesn't work for profiles with adapted parameters.

BRouter provides an undocumented way to [pass profile parameters as parameters of the request](https://github.com/abrensch/brouter/blob/master/brouter-server/src/main/java/btools/server/RouteServer.java#L216) which would offer the possibility for persistent urls even with adapted parameters.

* Default parameters: https://brouter.de/brouter?lonlats=8.723037,50.000491|8.712737,50.002899&nogos=&profile=trekking&alternativeidx=0&format=geojson
* Adapted parameter: https://brouter.de/brouter?lonlats=8.723037,50.000491|8.712737,50.002899&nogos=&profile=trekking&alternativeidx=0&format=geojson&profile:avoid_unsafe=1

nrenner:
Yes, that would be one of the tasks with some higher priority.

@abrensch [mentioned](https://github.com/nrenner/brouter-web/issues/223#issuecomment-517437902) a while ago that those are only a hack:

> So that' currently only a hack and profile:xxx does not overwrite params of the "StdModel", and there are other hacks, e.g. does profile:vmax at the same time requests the speed-limit-profile.
But if that should become a feature I could of course remove the hacks and make it more consistent.

What is the current state there?

zod:
It seems like it hasn't changed much. The special handling for `profile:vmax` was removed, but as it seems overwriting parameters of `StdModel` isn't possible. I'm not sure if we need to overwrite those parameters or it's sufficient to overwrite parameters which are provided by the profiles.

abrensch:
>> or it's sufficient to overwrite parameters which are provided by the profiles

subtile difference between "provided by/to the profiles"

The way I did it in KinematicModel is to inject the url-parameter when the Model queries a parameter from the profile (If you look at https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/StdModel.java here the parameter keyValues of the init method is ignored) 

But that means it does not work for parameters which are used BY the profile.

So we would need to inject the parameter before the profile is parsed, same way/same syntax as BRouter-Web already does it.

[...]

Issue #472, opened 2022-01-17
Export/import profiles


EssBee59:
Hello,

Using the Brouter-web users can choose a profile among a long list of available standard profiles, AND have the possibility  to „customize“ an existing profile.
(as example the parameter turninstructionmode, consider_elevation, considerTurnRestrictions  can / should be adapted in many situations)

Currently (and without registration) a modified profile can not be saved and later resused in an easy way.

I think, an extention to support « save profile» and « restore profile» would be nice to have !

As example 2 new functions could should be implemented (with a limited workload):
-	Export (download !) the current active profile (as modified by the user)
-	Import (upload and activate) a profile stored locally

Also user of the Brouter-app could use this new feature:
Brouter-app users generally developpe and test new profiles with the Brouter-web. The downloaded „brf“ file could then be easyly forwoarded to the smartphone (as mail attachement or with an other tool).
On the smartphone side, a click on the attachement allows a very simply "semi-automatic" import of the profile into the Brouter-app (feature available in Brouter-app from rel. 1.6.3)

EssBee59:
No comment to this suggested feature after 27 days...

Before closing it (or developping a prototype!), I liked to get a general opinion of the developers:
-Feature not of interest?
(yes, I try to make the Brouter-web more confortable for non experienced users, not for IT experts)
-Is not secured?
Possibly, but I do not see a real risk, as every users remain responsible for his changed profiles)
-Workload too high?

Thank in advance for any comment.
If interest is found, I could start to create a prototype.
Additionally to the described feature above, I think about saving a default-profile name as cookie: This allowing to activate this default on start if the user decides/configures  so.

EssBee59:
Hello Markus,

Thank a lot for all the informations in your comment!

My request is coming from bike-friends who are not familiar with cut/paste and liked to have a saving function for modified profiles.
- generally they are modifiing only few parameters
- usage of brouter-web on the PC (with big screen), not on the smartphone where only brouter-app is in use

I do not know the feature "Storing profiles in the browser itself", but the storage area is not important, the goal is to ability to reuse in an easy way a private cutom-profile.

FFMbyBicycle:
There are already some discussions about this "sharing"-function out there and I miss this feature a lot, too :-) I think #26 is the main issue, maybe somebody should close the other ones to avoid confusion...

* #26
* #300
* #299
* #232
* #149
* [...]

EssBee59:
Hello FFMbyBicycle,

[...]

Issue #26, opened 2015-08-02
Permalinks with custom profiles


poutnikl:
Would be possible to refer the cached profile in Permalink similarly as in CSV URL,
and eventually fallback to Trekking one when cache expires ?

BTW, what is cache retention of custom profile ?

I have noticed the CSV download links contains URL reference to used cached custom profile

http://h2096617.stratoserver.net:443/brouter?lonlats=17.125454,48.843028|17.215233,48.842463&nogos=&profile=custom_1438472332376&alternativeidx=0&format=gpx

But Permalinks do not.

http://brouter.de/brouter-web/#zoom=13&lat=48.8389&lon=17.1572&layer=OpenStreetMap&lonlats=17.125454,48.843028|17.215233,48.842463&nogos=&alternativeidx=0&format=geojson

I suppose reason is obvious - The profile is cached just temporarily and will not be permanently available to Permalink. But the drawback for that is, that **Permalink presented route does not follow the custom profile, but default Trekking profile, so author and recipient of Permalink will see 2 different routes.**

matkoniecz:
duplicate of #9?

poutnikl:
Partially. as it would address temporary status of the profiles.
But when implemented both issues will be addressed.

nrenner:
The custom profile is not included in the Permalink on purpose. The point is rather not to take responsibility for user content. And a permanent link on temporary content would give the wrong impression.

I don't know what the cache retention of custom profiles is, the server is managed by Arndt.

Sharing a Permalink with a custom profile active is indeed misleading and a drawback right now. I hope this will be either solved with #9, or there should at least be a warning - e.g. to first export the profile - so I leave this issue open.

poutnikl:
The intention was to have Permalink right temporarily e.g. for online discussion with people not aware about such things like custom profiles. I just guess the profile is cached while I am working with it, and if during that that time I provide the link.

:-) Perhaps, ther should be available another link - "TempolinkL"   :-)

But I guess incoming promised profile features will address much of that by their own way.
Perhaps saved and shared profiled will be able to be referenced in Permalinks in 0.6.0+ versions.
I am looking forward to it.

HankyPanky19:
Hi,
there is no profile with
consider_elevation=false
stick_to_cycleroutes=true
Since I often use that setting I would prefer to have "any way" to change profile at the beginning of a session. I'm tired to change and upload the "trecking" profile again and again at session startup...

[...]

PR #706, opened 2023-02-25
[WIP] [RFC] Include profile options in hash

tbsmark86:
Hi!

This is an Idea to address the various issues around this topic. For instance #26 #666 maybe #472
The Idea is to extend the &profile=fastbike value to be itself a nested parameter list. I'am doing this by replacing it with an URI.

The current minimal prototype will create a url like this:
http://localhost:3000/#map=12/51.8751/12.3835/osm-mapnik-german_style,route-quality&lonlats=12.461743,51.882955;12.416908,51.840657&profile=bw%3Afastbike%3Fallow_ferries%3Dfalse
Which when when loaded again will still have the allow_ferries set to false.

I've planed four different formats:
* `<profile-name>` -  just like it is currently when options are not required
* `bw:<profile-name>?<arg>=<value>` - for a standard profile with options
  It's an url with a custom scheme - 'bw:'
* `bwl:local` - for a manually edit profile
  Intention is to show a Message to anyone using the url that the profile is not available
* `http://example.com/profile.brf - for an external profile`
  This is for a future extension: Allowing the user to add custom profiles like they can add custom layers

**Is there any interest in this approach?**

Code wise I've started with creating a new class (ProfileData) to have more centralized state for all this profile related data.
My Intention is to move more of the current processing to this class so the UI-Elements can focus more on the UI part. Hopefully this would allow me to add some unit-tests for this stuff.
Of course this will require a lot of refactoring and likely new bugs. Also it will take quite some time.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -75,6 +75,7 @@ var paths = {
             'js/Util.js',
             'js/Map.js',
             'js/LayersConfig.js',
+            'js/ProfileData.js',
             'js/router/BRouter.js',
             'js/util/*.js',
             'js/format/*.js',
--- a/js/ProfileData.js
+++ b/js/ProfileData.js
@@ -1,0 +1,153 @@
+/* Central storage for the currently selected profile.
+ *
+ * Idea is to collect all state concerning profile here without beeing tied to
+ * a specific UI element.
+ *
+ * The current profile is serialized as a profile ID or as an URL.
+ * 1) profile ID: like the profile name on the server
+ * 2) As URL "bw:<id>?opt=value": Default profile with user defined options
+ * 3) As URL "http://example.com/profile.brf?opt=value": Profile stored on external location
+ *     possibly with user defined options
+ *     TODO: NYI
+ * 4) As URL "bwl:local": Marks a manually user edited profile.
+ *     The last state of this is stored an can be accessed again
+ *     TODO: NYI
+ */
+BR.ProfileData = L.Evented.extend({
+    // Currently active profile name/id for backend Request
+    // Possible values:
+    // - any known profile name for backend
+    // - L.BRouter.CUSTOM_PREFIX - not-yet-uploaded custom profile
+    // - L.BRouter.CUSTOM_PREFIX+_1234 - uploaded custom profile
+    id: '',
+    // Original profile name/id before user made custom changes
+    // Or URL of external Profile
+    baseId: '',
+    // Options the user changed in the profile (query-string)
+    options: '',
+    // Profile is unchanged
+    isDefault: false,
+    // Profile is loaded from external URL
+    isExternal: false,
+
+    // SourceCode of current profile. Private!
+    _profileText: '',
+
+    _lastInput: false,
+
+    initialize: function () {
+        this.selectProfile(null);
+    },
+
+    /** Set selected Profile */
+    selectProfile: function (input) {
+        if (this._lastInput === input) {
+            // break event loops
+            return;
+        } else if (input === null) {
+            // set default
+            input = BR.conf.profiles[0];
+        }
+        // assert input
+        if (typeof input !== 'string') {
+            throw new TypeError('Invalid argument');
+        }
+        this._lastInput = input;
+        try {
+            this._parseProfileName(input);
+        } catch (e) {
+            console.error('Invalid Profile. Reset to Default', e, input);
+            this._parseProfileName(BR.conf.profiles[0]);
+        }
+        this.fire('changed');
+    },
+
+    _parseProfileName: function (input) {
+        if (BR.conf.profilesRename?.[input]) {
+            input = BR.conf.profilesRename[input];
+        }
+        this.isDefault = BR.conf.profiles.includes(input);
+        this.isExternal = false;
+        if (this.isDefault) {
+            this.id = this.baseId = input;
+            this.options = '';
+        } else {
+            let url = new URL(input);
+
+            // split hash and url
+            this.options = url.search;
+            url.search = '';
+
+            if (url.protocol === 'bw:') {
+                // default profile with custom options
+                if (BR.conf.profiles.includes(url.pathname)) {
+                    this.baseId = url.pathname;
+                    if (this.options) {
+                        this.id = L.BRouter.CUSTOM_PREFIX;
+                    } else {
+                        this.id = this.baseId;
+                        this.isDefault = true;
+                    }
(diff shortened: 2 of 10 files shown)


nrenner (commented):
Thanks for the suggestion!

As I understand, this would be a pure client-side solution that still relies on modifying and uploading a custom profile?

However, the server now provides a way (according to #666) to pass profile parameters that override the corresponding variable of an unmodified standard profile, so the goal should rather be to get rid of using a custom profile for profile options and pass those as URL parameters to the server instead.

The server forwards URL parameters with a [`profile:` prefix](https://github.com/abrensch/brouter/blob/b735cd3e4ea1d4f25b37bdfa2c1384ad24cba69b/brouter-server/src/main/java/btools/server/RouteServer.java#L200) to be set in the [`RoutingContext`](https://github.com/afischerdev/brouter/blob/6d64da3a84d56bab2acfeff5041d5f61017f9bec/brouter-core/src/main/java/btools/router/RoutingContext.java#L114-L121). Although not using the latest master state, this already seems to work:
- https://brouter.de/brouter?lonlats=12.461743,51.882955|12.416908,51.840657&alternativeidx=0&format=geojson&profile=trekking&profile:allow_ferries=0  
(view in [geojson.io](http://geojson.io/#data=data:text/x-url,https%3A%2F%2Fbrouter.de%2Fbrouter%3Flonlats%3D12.461743%2C51.882955%7C12.416908%2C51.840657%26alternativeidx%3D0%26format%3Dgeojson%26profile%3Dtrekking%26profile%3Aallow_ferries%3D0))

I would expect the client URL to use a corresponding format, e.g. `&profile=trekking&profile:allow_ferries=0`.

> - http://example.com/profile.brf - for an external profile
>   This is for a future extension: Allowing the user to add custom profiles like they can add custom layers

abrensch and I agreed not to automatically load and display content from arbitrary external URLs. Instead, the decision was to offer a way to make a profile permanent by uploading it to a separate storage location that doesn't get cleaned, e.g. using a "share" button, and then to include the shared profile name for that in the URL. A first step for that was already made ([shared profile dir](https://github.com/abrensch/brouter/commit/b154851f1fd46e4c205ab15e0d669e39baf6142a) commit), but still missing the upload and client part.

So I see #26 and #666 as two different things. #472 would be just some manual File operations that load/save to/from the profile editor, maybe with an additional URL field like for loading no-go areas, but not modifying the URL. Uploading would still be an additional manual step, just like with pasting or editing.

tbsmark86:
> As I understand, this would be a pure client-side solution that still relies on modifying and uploading a custom profile?
Yes.

> However, the server now provides a way (according to https://github.com/nrenner/brouter-web/issues/666) to
Well that reads like this was not really intended by the Server? 
I'am not aware of any api doc for the server?

> abrensch and I agreed not to automatically load and display content from arbitrary external URLs.
Yes. My Idea was to first ask the user.
fgcvfg
Overall I read that as a 'no' to my question?

Or this profile:xyz is the 'official' way for the backend and just drop the  profile-is-a-nested-url part?

nrenner (changes requested):
> Well that reads like this was not really intended by the Server?

That's not my understanding, I always found that a useful feature and the way to go. While it started as a quick hack for a prototype, recent changes, that are part of a larger, still ongoing update, apparently turned this into something proper.

> I'am not aware of any api doc for the server?

General reference from the [readme](https://github.com/abrensch/brouter#run-the-brouter-http-server):
> The API endpoints exposed by this HTTP server are documented in the [brouter-server/src/main/java/btools/server/request/ServerHandler.java](https://github.com/abrensch/brouter/blob/master/brouter-server/src/main/java/btools/server/request/ServerHandler.java) file.

But as zod mentions in #666, the `profile:` handling is not documented for HTTP yet, but the Android interface has a [`extraParams`](https://github.com/abrensch/brouter/blob/b735cd3e4ea1d4f25b37bdfa2c1384ad24cba69b/brouter-routing-app/src/main/aidl/btools/routingapp/IBRouterService.aidl#L33) property for that.

>> abrensch and I agreed not to automatically load and display content from arbitrary external URLs.

> Yes. My Idea was to first ask the user.

It might not be obvious from the external URL in a shared link alone if it loads offensive or malicious content or not, until the users actually load and see it. The reasoning was, that by only sharing content from our server, we can control and potentially delete what gets shared.

> Overall I read that as a 'no' to my question?

To the approach as proposed, yes, I'm afraid, not to the general idea of including profile options in the hash. Having a separate ProfileData class might still be useful though.

> Or this profile:xyz is the 'official' way for the backend and just drop the profile-is-a-nested-url part?

I strongly think that passing profile:xyz parameters to the server instead of a custom profile is the way to go for profile options. By using it, we would make it official. If there is more work required on the server it would be part of the job or a feature request for the server. But from my understanding of the discussion in #666 this is ready to use.

Ideally, the URL request to the server would be reflected in the same way in the client URL hash, instead of profile-is-a-nested-url.

So if you're still interested, my suggestion would be to limit the PR to profile options, using profile:xyz server parameters instead of a custom profile as suggested in #666 and to exclude profile sharing (#26) and loading (#472).

nrenner:
> Also feels more complex moving around multiple values instead of one opaque string I have think about that.

But that's only when converting to/from URL, internally it would be a single object with key/value properties?

I don't like encoded "blobs" and would prefer to keep the URL unencoded and readable whenever possible. An alternative to multiple `profile:*`-prefixed URL parameters could be a single parameter with the `key=value` pairs separated e.g. by semicolon (`;`) instead of the parameter seperator (`&`), like we do e.g. in the `lonlats` parameter. The URL structure should be the same with that sent to the server, so this would need to be discussed in a server issue.

> Also I think once this split of ProfileData class is in place it should be more simple to add localStorage solution for saving you 'personal' custom profile more easly. In a way this external url stuff would still be there you just have to copy&paste it once.

For me, using localStorage to remember a profile is less a technical but more a conceptional issue, how it fits with permanent/shared profiles on the server, or if localStorage is even needed once you can permanently store profiles on the server.

> General Question: The current code is lots of "callback hell". Is this simply because it's rather old or are there any reasons to avoid more modern Promise or async/await ?

Using callbacks is not automatically "hell". In my understanding that term refers specifically to multiple nested callbacks, which I don't think we have that many. Originally I intended to still support older browsers and Babel was only introduced very late in the project. Now we can use modern JS features for new stuff and also gradually refactor what is being worked on (but please in a separate commit).

Issue #677, opened 2022-11-18
Load server hosted GeoJSON layers as overlays


sjdellis:
Is there any easy way to integrate server hosted GeoJSON layers hosted on a server into the overlays?

I can added them (to the map only) with the leaflet-ajax plugin:

`var geojsonLayer = new L.GeoJSON.AJAX("./data.json"),{
 	filter: function(feature, layer) {
  		 return (feature.properties.layer == "Route" && feature.properties.grade == "Grade 1");
  	 }
  });
  geojsonLayer.addTo(map);`

But it would be great to have the functionality built in and accessible by configuring layers.js with something like  "dataSource": "json" ?  or is this already in there and I've just not understood?

PR #727, opened 2023-04-20
Support remote GeoJSON layers

zod:
I've tried to achieve similar things as #677 and wanted to add a remote GeoJSON as as layer. Currently this seems to be possible using maplibre with a style file which uses a [geojson source](https://maplibre.org/maplibre-style-spec/sources/#geojson). It feels a bit sluggish and comments on leaflet-maplibre mention [performance issues](https://github.com/maplibre/maplibre-gl-leaflet#motivation).

Leaflet also provides GeoJSON support, but it seems there is currently no way to specify them in layer configuration.

I've implemented a minimal version which also support the [simplestyle-spec](https://github.com/mapbox/simplestyle-spec).

To test it you can add a layer configuration which uses the `example.json` from `simplestyle-spec`
```json
{
    "geometry": null,
    "properties": {
        "id": "geojson-simplestyle",
        "overlay": true,
        "type": "geojson",
        "url": "https://raw.githubusercontent.com/mapbox/simplestyle-spec/master/1.1.0/example.geojson"
    },
    "type": "Feature"
}
```

Because I'm not familiar with JS development there's a pending issue: Husky warned me because `fetch` isn't available in all supported browsers. I thought babel & polyfills should take care of that...

### DIFF
--- a/js/LayersConfig.js
+++ b/js/LayersConfig.js
@@ -349,7 +349,7 @@ BR.LayersConfig = L.Class.extend({
     },
 
     createGeoJsonLayer: function (props) {
-        const layer = L.geoJSON(undefined);
+        const layer = L.geoJSON(undefined, BR.Track.getGeoJsonOptions());
         fetch(props.url).then(async (response) => {
             const geojson = await response.json();
             layer.addData(geojson);
--- a/js/LayersConfig.js
+++ b/js/LayersConfig.js
@@ -348,6 +348,15 @@ BR.LayersConfig = L.Class.extend({
         }
     },
 
+    createGeoJsonLayer: function (props) {
+        const layer = L.geoJSON(undefined);
+        fetch(props.url).then(async (response) => {
+            const geojson = await response.json();
+            layer.addData(geojson);
+        });
+        return layer;
+    },
+
     createLayer: function (layerData) {
         var props = layerData.properties;
         var url = props.url;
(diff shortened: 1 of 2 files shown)


nrenner (commented):
Thanks for the PR!

Nice that you added the simplestyle support to the `Track.getGeoJsonOptions` util function. This method currently [filters out points](https://github.com/nrenner/brouter-web/blob/79a8c6ab9afdaf2083f7542e35d5823f2ed133c4/js/util/Track.js#L22-L25), because they're added as POI markers on track load. But I guess we would rather want normal markers here. I will do some refactoring after merging and move the filter.

We currently add polyfilled features manually to [eslint config](https://github.com/nrenner/brouter-web/blob/79a8c6ab9afdaf2083f7542e35d5823f2ed133c4/.eslintrc.json#L8), as a reminder to check if already included (don't know if there is an automatic way). And `fetch` is actually [missing in core-js](https://github.com/zloirock/core-js#missing-polyfills), the polyfill lib we use. We have some current inconsistencies and past changes I wonder about, so I will also address this afterwards.

Is this ready to merge, apart from the above?

zod:
I didn't add support for markers because it seems Leaflet doesn't support specifying colors by default

Other than the issue described above it's ready (and I've removed the draft status).

nrenner:
Thanks for your contribution!

PR #729, opened 2023-05-06
Share Dialog with QR Code

mjaschen:
This pull-request implements sharing a route by QRCode, external services or by using the clipboard

- add a new button "Share" to the main navigation
- when clicked, a new modal dialog opens and shows share options for the planned route, which are currently:
  - a button to copy the URL of the current route to the clipboard
  - share buttons for 3rd party services, such as Facebook, Whatsapp, Email (and Mastodon, but currently disabled via config[^1])
  - a QRCode
- the QRCode is a representation of the current route (i.e. the current URL) with an added query parameter to the URL: `?export=dialog`
- when BRouter-Web is loaded with the `export=dialog` query parameter, the export dialog is shown instantly (as discussed in #370)
  - using a query parameter (`export`) and a value (`dialog`) instead of only a paramater is required, because `URLSearchParams` needs both when adding a new parameter
  - we've the opportunity to add further options in the future (`download`, `gpx` etc.)
- add a new keyboard shortcut 'A' (“action”) which opens the share dialog (would have used 'S', but that's already taken)
- add 3rd-party QRCode library
- add `URLSearchParams` polyfill for ancient browsers as Husky didn't allow me to commit otherwise …
- add license information to README
- TODO: Translations

[^1]: Mastodon instances respond with HTTP 502 when the submitted text (= the route URL in our case) is longer than the maximum post character limit on that instance. That will probably confuse users. I decided to keep the code here as this bug will probably be fixed in the future.

### DIFF
--- a/js/control/ShareRoute.js
+++ b/js/control/ShareRoute.js
@@ -16,18 +16,6 @@ BR.ShareRoute = L.Class.extend({
     initialize: function () {
         L.DomUtil.get('shareButton').onclick = L.bind(this.share, this);
         L.DomEvent.addListener(document, 'keydown', this._keydownListener, this);
-
-        this.shareButton = $('#shareButton');
-        this.update([]);
-    },
-
-    update: function (latLngs) {
-        if (latLngs.length < 2) {
-            this.shareButton.addClass('disabled');
-            return;
-        }
-
-        this.shareButton.removeClass('disabled');
     },
 
     share: function (event) {
--- a/js/control/ShareRoute.js
+++ b/js/control/ShareRoute.js
@@ -16,6 +16,18 @@ BR.ShareRoute = L.Class.extend({
     initialize: function () {
         L.DomUtil.get('shareButton').onclick = L.bind(this.share, this);
         L.DomEvent.addListener(document, 'keydown', this._keydownListener, this);
+
+        this.shareButton = $('#shareButton');
+        this.update([]);
+    },
+
+    update: function (latLngs) {
+        if (latLngs.length < 2) {
+            this.shareButton.addClass('disabled');
+            return;
+        }
+
+        this.shareButton.removeClass('disabled');
     },
 
     share: function (event) {
--- a/js/control/ShareRoute.js
+++ b/js/control/ShareRoute.js
@@ -52,11 +52,12 @@ BR.ShareRoute = L.Class.extend({
     },
 
     /**
-     * Renders QR Code for the currect route:
+     * Renders QR Code for the current route:
      *
-     * - hides the export modal dialog (QR Code is shown in a new dialog)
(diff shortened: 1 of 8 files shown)


js/control/ShareRoute.js
```diff
@@ -0,0 +1,149 @@
...
+                    if (mastodonServer.indexOf('http') !== 0) {
+                        mastodonServer = 'https://' + mastodonServer;
+                    }
+                    window.open(mastodonServer + '/share?text=' + encodeURIComponent(self.getShareUrl()), '_blank');
+                });
+        }
+    },
+
+    /**
+     * Renders QR Code for the currect route:
```
nrenner, line 55: currect -> current?

js/control/ShareRoute.js
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * - hides the export modal dialog (QR Code is shown in a new dialog)
```
nrenner, line 57: doesn't apply to current implementation anymore?

nrenner (changes requested):
Thanks for the PR, nice!

The `url-search-params-polyfill` isn't needed, as a `URLSearchParams` polyfill is already [included in core-js](https://github.com/zloirock/core-js#url-and-urlsearchparams). The added entry in `.eslintrc.json` is enough in this case, to make it stop complaining. That's just a manual checklist, it can't automatically detect missing polyfills.

There is an error on download when sharing a single route point, I would suggest to add an additional check before opening the export dialog, probably like `if (routing.getWaypoints().length < 2)` as in [`Export.update`](https://github.com/nrenner/brouter-web/blob/c715d2bab1305264652c54cceb1f8ed6436ccbf0/js/control/Export.js#L45).

js/control/ShareRoute.js
mjaschen: good catch ;-)

js/control/ShareRoute.js
mjaschen: Yes, of course. I've updated this doc block.

mjaschen:
I've added a check so that the button is enabled only if at least 2 route points are set.

nrenner:
> I've added a check so that the button is enabled only if at least 2 route points are set.

No strong opinions, we can merge as is, but there might be use cases where it makes sense to only share the current map view or a single route point. Therefore I meant to still allow sharing, just not to open the export dialog at the receiving end by adding that check to [index.js](https://github.com/nrenner/brouter-web/pull/729/files#diff-3f689a613b62130589affae189bdfbcff8533286dcf50c288ccf61bd9b59c11cR319).

Don't know if my comment was misleading or if you prefer to disable the share dialog instead?

nrenner:
Thanks!

PR #733, opened 2023-06-02
Wrap long URLs in custom layers overlay

mjaschen:
Long URLs were overflowing in the custom layers dialog. This pull request adds a class attribute to the cells of the URL column and uses CSS to wrap those URLs. Additionally, the URLs are printed in a monospace font, but that's for aesthetics only.


mjaschen:
Before:

overlays-bug

After:

image

nrenner (commented):
Thanks!

PR #735, opened 2023-06-11
Add "Help" dialog which shows available keyboard shortcuts

mjaschen:
Dialog can be toggled via button in "About" or via keyboard shortcut "?"

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -527,6 +527,46 @@ button.deactivate-beeline-active.disabled {
     margin: -6px 0 6px 20px;
 }
 
+/* help dialog */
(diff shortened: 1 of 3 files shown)


nrenner (commented):
The dialog looks nice!

I wondered if we should reuse the tooltip texts as descriptions in order to avoid duplicate translations of similar texts, but it would probably be a lot of work gathering those and making them usable for both, and there still might be reasons to differ.

It's a bit inconsistent that we're using uppercase letters in the tooltips and lowercase in the help dialog, but I don't really mind.

PR #753, opened 2023-07-12
Adapt voice hints and exporting to changes in BRouter 1.7.X

rkflx:
Patches contained in this series (see commit messages for details), all relating to adapting to changes in BRouter 1.7.0:
- Add BL and TLU to VoiceHints and adapt indexing to BRouter 1.7.0
- Emit console warning when using an unsupported BRouter version
- Document new export formats added in BRouter 1.7.0 and fix export error
- Prevent Gpsies-style export from throwing for undefined symbols

Further changes would be needed (e.g. new Locus format, straight line voice hints, 180° u-turn voice hints at cul-de-sac-style vias, height calculation buffers etc.), but that's all there is for now.

Regarding the new console warning if BRouter is too old, I'm open to better suggestions, of course. I haven't added this new requirement to the [CHANGELOG](https://github.com/nrenner/brouter-web/blob/master/CHANGELOG.md) yet, but it should probably be mentioned there and in the release notes once a new release is tagged. Not sure how increasing the required BRouter version was handled in the past.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,3 +148,5 @@ Copyright (c) 2018 Norbert Renner and [contributors](https://github.com/nrenner/
     Copyright (c) 2012 davidshimjs [The MIT License](https://github.com/llyys/qrcodejs/blob/master/LICENSE)
 -   [Bootstrap Icons](https://github.com/twbs/icons)
     Copyright (c) 2019-2023 The Bootstrap Authors [The MIT License](https://github.com/twbs/icons/blob/main/LICENSE)
+-   [compare-versions](https://github.com/omichelsen/compare-versions)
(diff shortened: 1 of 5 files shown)


nrenner:
Thanks for the PR! I'm fine with merging that as is.

> Regarding the new console warning if BRouter is too old, I'm open to better suggestions, of course.

Fine by me.

> Not sure how increasing the required BRouter version was handled in the past.

Not at all, other than providing a working standalone bundle with each release. On my to-do list is to include brouter as a dev or peer dependency, which initially wasn't possible because of the GPL license.

nrenner:
Thanks!

PR #777, opened 2023-10-07
Enable sharing via Mastodon

mjaschen:
Sharing works now with newer Mastodon releases.

SCR-20231007-mjtx

SCR-20231007-mkaj


mjaschen:
Included in #785

Issue #775, opened 2023-09-26
Extend support for custom layers


zod:
Currently it's only possible to add TMS or Overpass queries as custom layers. Some map tiles are only available as vector tiles (MVT) which brouter-web supports, but not for custom layers.

I made a simple mockup how the extended dialog could look like
![image](https://github.com/nrenner/brouter-web/assets/49181/180dd1ac-b147-4ec0-9c56-fb0a33bab5b0)

For some types (Overpass & GeoJSON) the overlay checkbox would always be checked, because they don't make any sense as baselayer. We could also allow to specify additional information based on the type (e.g. maxzoom for raster layers).

PR #780, opened 2023-10-11
Store custom layers as FeatureCollection

zod:
Replace custom format with GeoJSON Features which are already used for layer configuration. This allows storing also MVT or GeoJSON layers.

Currently custom layers are stored in localStorage in a custom format which supports only TMS, OverpassAPI and OpenStreetMapNotesAPI(?). For #775 support for additional formats is required. Instead of extending the custom format this reuses the existing handling from LayerConfig which uses [JOSM's map format](https://josm.openstreetmap.de/wiki/Maps) serialized to GeoJSON.

Existing localStorage are converted to the new format.


bagage:
Thanks @zod for your contribution!

PR #643, opened 2022-09-16
Add IGNF Scan25 baselayer (requires key)

sylvainavelo:
Adds the IGN France - Carte topographique (Scan25) base layer.

This layer requires a valid key for "ignf" (see key.template.js in the root directory)

Key can be generated by:
* creating an account on https://geoservices.ign.fr/ (website is entirely in french)
* once logged in, visit "Mon Espace" > "Mes clés de service web" (https://geoservices.ign.fr/mes-cles)
* choose "Créer une clé"
* choose the first license "Usage gratuit" (free use)
* copy and use your new key
* optionally (recommended in production), add a referer safety on your key, so that only your website can send requests using it.

----

Related issue : #463

PR #786, opened 2024-01-02
Change IGNF layers URLs

sylvainavelo:
Change urls from wxs.ign.fr to data.geopf.fr for IGNF layers

The old urls will stop working on March 15th 2024 (according to https://geoservices.ign.fr/bascule-vers-la-geoplateforme in french)

There is a publicly available sample `apikey` parameter for testing the *ignf-scan25* layer: '`ign_scan_ws`' (https://geoservices.ign.fr/actualites/2023-11-20-acces-donnesnonlibres-gpf in french)

These changes maintain the layers added by #636 and #643

### DIFF
--- a/keys.template.js
+++ b/keys.template.js
@@ -20,7 +20,7 @@
         // Mapillary, https://www.mapillary.com/dashboard/developers
         mapillary: ``,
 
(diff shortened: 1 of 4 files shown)


bagage:
Hello @sylvainavelo, sorry for the late reply, thanks for the update!

PR #819, opened 2024-07-02
add 'break' to switch to prevent fall-through

mjaschen:
A missing break lead to fall-through in the switch statement.


bagage:
Good catch! :D

PR #822, opened 2024-09-22
add overpass query for railway station

gorgobacka:
This PR adds the Overpass query for railway stations to the transportation layers.

I'm adding the overpass query regularly, because it's currently missing and I think, it should be available by default.


gorgobacka:
@bagage Can you please check this PR?

Issue #792, opened 2024-03-11
Update Leaflet to 1.9.4 to fix scroll wheel issue (zoom by 2 on Firefox) 


Nekzuris:
Your current Leaflet version still has a bug where the scroll wheel zoom by 2 levels on Firefox.
https://github.com/Leaflet/Leaflet/issues/7403

gorgobacka:
It seems that there was an update to 1.9.2 some time ago but it was reverted due to some errors: https://github.com/nrenner/brouter-web/commit/dd0715076ac449b1256c18a11403c7f45e540398

Maybe there are some more changes required to update the current Leaflet version.

gorgobacka:
I worked on it this evening. At least I got it working. However, there is still an exception in the browser that I haven't solved so far.

I created a draft PR #824 . Maybe someone has a idea.

PR #824, opened 2024-09-23
Update leaflet to 1.9.4

gorgobacka:
This is a starting point to update Leaflet to the most recent version (solves #792).

I managed to fix the most obvious issue with the no-go-area. They are usable.

However, there is still one exception on the browser console that I couldn't solve so far.
leaflet-nogo-areas

Maybe someone else is able to fix it.

### DIFF
--- a/js/plugin/NogoAreas.js
+++ b/js/plugin/NogoAreas.js
@@ -96,7 +96,7 @@ BR.NogoAreas = L.Control.extend({
         });
 
         // prevent instant re-activate when turning off button by both Pointer and Click
-        // events firing in Chrome mobile while L.Map.Tap enabled for circle drawing
+        // events firing in Chrome mobile while L.Map.TapHold enabled for circle drawing
         L.DomEvent.addListener(this.button.button, 'pointerdown', L.DomEvent.stop);
 
         L.DomEvent.addListener(document, 'keydown', this._keydownListener, this);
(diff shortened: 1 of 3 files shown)


mjaschen:
I think I've figured it out and made a pull request in your repository.

It'd be nice if you can verify that it works now and update this PR.

https://github.com/gorgobacka/brouter-web/pull/1

gorgobacka:
@mjaschen found the missing piece. :+1: 

This PR is ready now.

mjaschen:
Unfortunately there's a problem which breaks the “No Go Area” feature on touch devices.

*tap and hold* to create a circle doesn't work.

I've spent some time on the weekend on the search for a fix and didn't found one. But at least I've a workaround which allows placing fixed size no go areas with a single tap on touch devices, so it's usable at least.

Also, the routing library issues a call to `DomEvent.fakeStop()` when tapping the route, but that method no longer exists. That's another thing which has to be fixed before merging this PR. My workaround for now is to simply remove that method call: https://github.com/mjaschen/leaflet-routing/commit/0690d5e29a8156121f4f6d102bf380459ef01b5a

With these two workarounds, BRouter-Web works fine with Leaflet 1.9.4 – as far as I've tested it.

If I don't find any further bugs, I'll deploy it in the next days to bikerouter.de and create pull requests here. Maybe someone has an idea how to create proper bugfixes instead of workarounds.

bagage:
I don't have much time to test but I am more than happy to merge this PR when it's good enough to go live :) It'd be nice to fix the detected issues, but if it's really hard I think we can assess if that's really important (the nogo circle workaround seems good enough to me for instance).

If there are other testers available to give it a try and report issues, it would greatly help! Let us know @mjaschen when your instance is live with this branch so that it's easier to test it :)

And many thanks for the PR and follow-up!

mjaschen:
A version with Leaflet 1.9.4 was deployed to <https://bikerouter.de>

It contains this pull request with the following additional changes:

- “No Go Areas“ are placed with a single tap on touch devices; behavior on devices with mouse pointer is unchanged (click + drag). The tooltip adjusts accordingly. The initial size of the circle depends on the viewport width.
  - Todo: figure out how to use tap + drag on iOS touch devices
- Leaflet-Routing library ([diff](https://github.com/mjaschen/leaflet-routing/compare/2ad0176c72e32246d640759966a6e631fcd84b55..31e202b4b72bc7c407f07dd961134209297c883d)):
  - replace `fakeStop()` call on `DomEvent` as this method no longer exists
  - force showing the trailing line in draw mode after the most recent waypoint was deleted
  - Todo: check if there's an easier/cleaner way to achieve this

gorgobacka:
@mjaschen Have you received any feedback on bikerouter regarding this implementation during the last months? And is it worth to use it as it is?

PR #828, opened 2024-10-14
allow additional GeoJSON MIME type

mjaschen:
The MIME type for GeoJSON registered with IANA is `application/geo+json`, replacing the old value `application/vnd.geo+json`. The change was published with RFC 7946 in 2016.

BRouter reponses changed to `application/geo+json` in recent versions.

With this commit BRouter-Web accepts either of both MIME type strings in BRouter responses.


bagage:
Thanks! :+1:

PR #829, opened 2024-10-14
refact: use shorthand method and property syntax

simon04:
Ref: https://ui.dev/shorthand-properties


bagage (approved):

bagage:
Thanks @simon04 for the cleanup! Do you know if there's a way to enforce this via a linter or similar tool?

PR #830, opened 2024-10-19
Add “Maximum Speed” to analysis sidebar

mjaschen:
It shows the distribution of maximum speeds for all ways on the current route (if that data is available, otherwise it’s summed up under “unknown”).

`maxspeed:forward` and `maxspeed:backward` is respected in conjunction with `reversedirection`.

Hovering/clicking table rows to highlight matching segments on the route work the identical to the other analysis tables.

Additionally, all tags in the analysis tab (way type, surface, smoothness) are translateable now. The values were added to `en.json`.

Some HTML is rendered with template literals now, instead of concatenating strings.

Variable declarations were changed from `var` to `const`/`let`.

### DIFF
--- a/js/control/TrackAnalysis.js
+++ b/js/control/TrackAnalysis.js
@@ -180,13 +180,19 @@ BR.TrackAnalysis = L.Class.extend({
                         case 'surface':
                         case 'smoothness':
                             if (typeof analysis[tagName][wayTagParts[1]] === 'undefined') {
-                                let formattedName = i18next.t([
-                                    'sidebar.analysis.data.' + tagName + '.' + wayTagParts[1],
-                                    wayTagParts[1],
-                                ]);
+                                let formattedName;
+
                                 if (tagName.indexOf('maxspeed') === 0) {
-                                    formattedName += ' km/h';
+                                    formattedName = i18next.t('sidebar.analysis.data.maxspeed', {
+                                        maxspeed: wayTagParts[1],
+                                    });
+                                } else {
+                                    formattedName = i18next.t([
+                                        'sidebar.analysis.data.' + tagName + '.' + wayTagParts[1],
+                                        wayTagParts[1],
+                                    ]);
                                 }
+
                                 analysis[tagName][wayTagParts[1]] = {
                                     formatted_name: formattedName,
                                     name: wayTagParts[1],
--- a/js/control/TrackAnalysis.js
+++ b/js/control/TrackAnalysis.js
@@ -75,14 +75,14 @@ BR.TrackAnalysis = L.Class.extend({
     /**
      * Everytime the track changes this method is called:
      *
-     * - calculate statistics (way type, surface, smoothness)
+     * - calculate statistics (way type, max speed, surface, smoothness)
      *   for the whole track
      * - renders statistics tables
      * - create event listeners which allow to hover/click a
      *   table row for highlighting matching track segments
      *
      * @param {Polyline} polyline
-     * @param {Array} segments
+     * @param {Array} segments route segments between waypoints
      */
     update(polyline, segments) {
         if (!this.active) {
@@ -105,7 +105,7 @@ BR.TrackAnalysis = L.Class.extend({
         this.trackPolyline = polyline;
         this.trackEdges = new BR.TrackEdges(segments);
 
-        var analysis = this.calcStats(polyline, segments);
+        const analysis = this.calcStats(polyline, segments);
 
         this.render(analysis);
 
@@ -132,6 +132,7 @@ BR.TrackAnalysis = L.Class.extend({
     calcStats(polyline, segments) {
         const analysis = {
             highway: {},
+            maxspeed: {},
             surface: {},
             smoothness: {},
         };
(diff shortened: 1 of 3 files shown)


js/control/TrackAnalysis.js
```diff
@@ -75,14 +75,14 @@ BR.TrackAnalysis = L.Class.extend({
...
-     * - calculate statistics (way type, surface, smoothness)
+     * - calculate statistics (way type, max speed, surface, smoothness)
      *   for the whole track
      * - renders statistics tables
      * - create event listeners which allow to hover/click a
      *   table row for highlighting matching track segments
      *
      * @param {Polyline} polyline
-     * @param {Array} segments
+     * @param {Array} segments route segments between waypoints
```
bagage, line 85: nit: `segments route segments`? Shouldn't it be `route segments` alone?

locales/en.json
```diff
@@ -285,10 +285,66 @@
   },
   "sidebar": {
     "analysis": {
+        "data": {
+            "highway": {
```
bagage, line 289: why max_speed is not present here?

js/control/TrackAnalysis.js
```diff
@@ -175,14 +176,19 @@ BR.TrackAnalysis = L.Class.extend({
...
+                        case 'maxspeed':
                         case 'surface':
                         case 'smoothness':
                             if (typeof analysis[tagName][wayTagParts[1]] === 'undefined') {
+                                let formattedName = i18next.t([
+                                    'sidebar.analysis.data.' + tagName + '.' + wayTagParts[1],
+                                    wayTagParts[1],
+                                ]);
+                                if (tagName.indexOf('maxspeed') === 0) {
+                                    formattedName += ' km/h';
```
bagage, line 188: nit: not possible to have it within the i18n content directly? 

Eg add `sidebar.analysis.data.maxspeed.XXX` and in English set to `%s km/h`?

js/control/TrackAnalysis.js
```diff
@@ -279,29 +302,35 @@ BR.TrackAnalysis = L.Class.extend({
...
                 if (!analysis[type].hasOwnProperty(name)) {
                     continue;
                 }
                 analysisSortable[type].push(analysis[type][name]);
             }
 
-            analysisSortable[type].sort(function (a, b) {
-                return b.distance - a.distance;
-            });
+            if (type === 'maxspeed') {
```
bagage, line 323: nit: probably would have used a switch case for this but not a blocker at all :)

bagage (approved):
Thanks for implementing this feature :+1: looks good to me only few minor nitpickings but good to go!

js/control/TrackAnalysis.js
mjaschen: The first `segments` refers to the variable name, the three following words are the description.

As per [JSDoc](https://jsdoc.app/tags-param) a hyphen should be inserted after the parameter name. I'll add this in a follow-up pull request.

locales/en.json
mjaschen: See line [347](https://github.com/nrenner/brouter-web/pull/830/files/50e85509ac9c11bfc5b7c2542931048eb9bcb0ea#diff-c9218bed23f0aeb93d7e1a7d5f6e1345fae20f4b8692b27fb97fd0f37b3d26d7R347)

js/control/TrackAnalysis.js
mjaschen: Yes, good catch. I'll change it in a follow-up PR.

js/control/TrackAnalysis.js
bagage: Ah yeah ok my bad, :+1:

js/control/TrackAnalysis.js
```diff
@@ -180,13 +180,19 @@ BR.TrackAnalysis = L.Class.extend({
...
-                                let formattedName = i18next.t([
-                                    'sidebar.analysis.data.' + tagName + '.' + wayTagParts[1],
-                                    wayTagParts[1],
-                                ]);
+                                let formattedName;
+
                                 if (tagName.indexOf('maxspeed') === 0) {
-                                    formattedName += ' km/h';
+                                    formattedName = i18next.t('sidebar.analysis.data.maxspeed', {
+                                        maxspeed: wayTagParts[1],
```
bagage, line 187: You could totally remove the `if` specific to maxSpeed, and instead have a fallback mecanism:

1. try to use the full key `sidebar.analysis.data.highway.footway`
2. if not available, try `sidebar.analysis.data.highway`
3. otherwise simply fallback to highway type eg `footway` here  

It should cover all cases while being agnostic of the kind of analysis (if there are more analysis coming in the future :smirk:)

```js
const highwayType = wayTagParts[1];
formattedName = i18next.t(
    [
         'sidebar.analysis.data.' + tagName + '.' + highwayType,
         'sidebar.analysis.data.' + tagName,
          highwayType
    ], 
    {highwayType: highwayType}
);
```

If a translation key needs the `highwayType` it can use it, otherwise it can just skip it and it should work just fine.

mjaschen: I spotted some more possibilities to simplify the translations for the analysis sidebar, but I'd rather create a separate PR as there're changes on places unrelated to this PR. 

Can you agree with this?

PR #838, opened 2024-11-20
Refactor: Use URLSearchParams instead of L.Util.template for query 

tbsmark86:
L.Util.template was missing url-escaping, while the values should never contain anything that needs escaping better safe then sorry.

---
Just a random bit I've stumbled over while trying to understand the code for something else.

### DIFF
--- a/js/router/BRouter.js
+++ b/js/router/BRouter.js
@@ -119,29 +119,47 @@ L.BRouter = L.Class.extend({
 
     getUrl(latLngs, beelineFlags, pois, circlego, format, trackname, exportWaypoints) {
         var urlParams = this.getUrlParams(latLngs, beelineFlags, pois, circlego, format);
-        var args = [];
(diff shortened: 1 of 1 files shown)


mjaschen:
Here is double-escaping going on, which has to be fixed before merging. The call to `encodeURIComponent()` in 

https://github.com/nrenner/brouter-web/blob/master/js/router/BRouter.js#L346

has to be removed.

Possible test case:

```js
test('handles POI names with special characters', () => {
            const latLngs = [L.latLng(49.488117, 8.467712)];
            const pois = [{ latlng: L.latLng(49.4882, 8.4678), name: 'Café & Restaurant' }];

            const url = router.getUrl(latLngs, null, pois);

            expect(url).toContain('pois=8.4678%2C49.4882%2CCaf%C3%A9+%26+Restaurant');
        });
```

Issue #434, opened 2021-06-30
Feature Request: Enhance height diagram indicator with information about marked distance/elevation gain


tstreibl:
It's already possible to nicely show the height diagram of a route and to select a region in the height diagram between a start/end marker (the selected region is used to automatically zoom into the map). During selection the actual data of the end marker is displayed (height and the distance from route start to the end marker) but the indicator does not use the additional information from the start marker.

Feature Request:  would it be possible to additionally display the distance between start/end markers and the accumulated elevation gain between start/end markers? The height diagram of google earth would be a good example for what i mean.

Benefit:  on a bike trip, especially through the mountains, you often try to estimate what distance you can travel easily in a day or a certain time. Clearly you want to take the accumulated height into account. For example when you have to go over several passes and elevation accumulates to 1600m height you want to travel only 80 km, whereas on plain terrain you easily could go 120 km. By marking the start/end point in the heigh diagram you could easily detect where on your route (end marker) you reach your personal heigth/distance limits and if by chance a good accomodation is nearby. On a hilly landscape it's often surprising how all the many ascents accumulate in the end. Certainly you can do those calculation by hand as well, but it's tedious.

rkflx:
Thanks for moving the suggestion to a separate issue.

While the elevation graph is provided by https://github.com/nrenner/leaflet.heightgraph, I guess we can track it here too and think a bit outside of it's "box".

For example, where to display statistics about the selected route segment, e.g. elevation gain, distance, travel time?
- One way would be to add a separate textbox to the elevation graph. Only the elevation graph would reflect the selection.
- We could also try to reuse the status bar for displaying statistics of the selection instead. This way the elevation graph would function as a general tool to select / filter parts of a route (*), and also the route and analysis sidebars could react to that, a potential segment list sidebar (see #430) could offer a button "Copy selection to new route" etc.

(*) Nice to have in this case: mark selection visually on track too, add close/unselect button, allow extending/shrinking/moving the selection with the mouse in the elevation graph, ...

tstreibl:
Those are great ideas! And many thanks for this great projekt and all the people making it happen!

internationils:
[...]

Issue #416, opened 2021-05-16
Height map colors can be misleading


Phyks:
Hi,

I've just discovered the height map coloring feature, which is really great! It's really useful, especially in areas such as the Alps http://brouter.de/brouter-web/#map=9/45.7800/5.2055/cyclosm,Waymarked_Trails-Cycling&lonlats=5.692114,45.935881;5.378945,45.865159.

I, however, tend to find it a bit misleading in the plains. See http://brouter.de/brouter-web/#map=13/49.8455/0.6686/cyclosm,Waymarked_Trails-Cycling&lonlats=0.647337,49.834438;0.660552,49.837983 for instance (flat, but orange conveys a meaning of "not so easy") and http://brouter.de/brouter-web/#map=11/46.8111/-2.1529/cyclosm,Waymarked_Trails-Cycling&lonlats=-2.132155,46.819795;-2.07779,46.786895 (basically flat, but seems like we have some roughs segments here!).

![Screenshot 2021-05-16 at 15-42-25 Client web BRouter](https://user-images.githubusercontent.com/3856586/118399417-56d5d480-b65d-11eb-9017-b13df0785e09.png)

![Screenshot 2021-05-16 at 15-42-49 Client web BRouter2](https://user-images.githubusercontent.com/3856586/118399433-62c19680-b65d-11eb-8207-294ac8665192.png)

Here are a few proposals for discussion:

* First, I'd rescale the color scheme, to have very light green for "-1 / 1%" and orange as first incline up level. It seems reasonable to me that the color palette is centered on "zero incline" (green = all right = no efforts) instead of orange (a bit sporty).
* Then, the Y axis of the height map can be confusing by over-stretching the inclines, which combined with the colors can make a route feel like we are in the Alps (while we are along the sea :). My proposal would be to enforce a minimum span on the Y axis (something along 50m span should be good I think). For reference, Komoot enforces a minimum span of 200m (which is too much for the diversity of BRouter users in my opinion).

Also, it seems the elevation segments are defined by OSM segments (which have no reality on the ground). It might be worth doing a pass on the elevation graph in the frontend to resplit it and recolor it accordingly. This can be seen in http://brouter.de/brouter-web/#map=11/46.8111/-2.1529/cyclosm,Waymarked_Trails-Cycling&lonlats=-2.132155,46.819795;-2.07779,46.786895

[...]

Issue #361, opened 2021-01-19
Common color scheme for Heightgraph and route incline coloring?


nrenner:
*From "OSM Android bikerouting" group by Pedro Varela:*

The heightgraph looks awesome!
Would it be a good idea to apply the colors seen in the Heightgraph to the route incline coloring, so that it would match?

![Screenshot](https://09133693475671011370.googlegroups.com/attach/caaab6b446b04/brouter01.png?part=0.1&view=1&vt=ANaJVrEnbSlbIGKAnM1wGyFmHDk9sUYC5qGWpFEFHSVf1dkJlNHd386vAyYycDe3n1208as28Gf37PyYIGcgqlV9e1T0Lem0p-AiYy8pHPsqEQ_aIJR2WuM)

nrenner:
Incline route coloring was improved in [#313 (commit)](https://github.com/nrenner/brouter-web/pull/313/commits/900f9252761bb817c573110b933a2e946054f38e), vs. [Heightgraph colors](https://github.com/nrenner/brouter-web/blob/dc8302479e8d09def1ecd93fc746f6f0d45b080b/js/plugin/Heightgraph.js#L13-L58).

rkflx:
A while ago I calculated which colors to use for the categories so they would match the route color coding:

```
categories -5 to 5:
#0000ff
#0044ff
#00ddff
#00ffbb
#00ff55
#00ff00
#55ff00
#bbff00
#ffdd00
#ff4400
#ff0000
```

(Calculation done by linear interpolation for each color channel of the [cost color coding gradient](https://github.com/nrenner/brouter-web/blob/900f9252761bb817c573110b933a2e946054f38e/js/plugin/RoutingPathQuality.js#L23-L29), mapped to the midpoint of each height graph color category. Later realized I simply should have used `getRGBForValue()` from Leaflet.Hotline.)

Sorry there are no screenshots at this point, typing this up from my notes without access to my dev setup.

[...]

PR #849, opened 2025-05-08
Update leaflet.heightgraph to fix marker hiding & more

### LINKED ISSUE #408: Route Quality coding
Hi The popup on the map with the gradient percentage is now hide behind  the track.
I think it is only a matter of changing the z-index to place it on top of a track?
Screenshot_2021-05-10 BRouter website
Regards Jan

tbsmark86:
fixes #408 and similar bug in the graph itself

includes https://github.com/nrenner/Leaflet.Heightgraph/pull/1 to display selection stats

Of course there is no need to switch the heightgraph repo to me. If instead you like to pull the changes into your heightgraph repo and let the package.json point there.


tbsmark86:
After diving into the height graph code i played around with it some more and also worked on #361 and maybe #416 - meaning syncing the color on the map with that of the graph. Downside is that is very colorful now because of course real road are rarely steady climbing:

Screenshot 2025-06-17 at 20-24-15 BRouter Web Client

* Someone more clever may be able to get a better smoothing int the elevation data
* As you can see in the screenshot i also changed the 'selection' behaviour to a 'zoom in' on the graph
* And as a bouns i've added surface and highway to the graph's tooltip

What you guy's are thinking about this?

There are further directions we could take this
* syncing the route summary with the selection of the graph (instead of showing those values in the tooltip) (#434)
* add an additional line on top to show the surface coloring
* or: always show the same color as the hotline (unless hotline shows nothing then show gradient)
* sync the tooltip of graph with the detailed table display highlight
Any more thoughts on these?

